PR #311, opened 2014-03-20
Update test_rosdep_installers.py

RaniaRho:
removed test cases using installer.unique() as it is not used in rosdep2


wjwwood:
Closing due to inactivity.

PR #312, opened 2014-03-21
Include sources and cache with rosdep.

trainman419:
This packages the default sources with rosdep, and updates the build system to build a cache for inclusion with distributions (sdist, sdist_dsc and bdist_deb).

The sources are installed to /etc/ros/rosdep/sources.list.d/ and the cache is installed to /var/cache/rosdep . The search path for the cache is modified appropriately. The cache is set to world-readable.

This means that an installation from debs or via pip should install the sources and cache, and allow rosdep to work out-of-the-box again. This also works when rosdep is installed into a virtualenv.

ROSDEP_PREFIX is now used when searching for sources and the cache. This allows users to use a custom cache, and allows system admins to relocate the cache as necessary.

Open items. Some of these are changes, some could use work down the road.
 * [ ] When installing from debs, the cache is always overwritten with the packaged version
 * [ ] `rosdep init` still pull sources from github rather than using the built-in default sources
 * [ ] `rosdep init` is no longer required by default. `rosdep update` should now be invoked with sudo. The public install docs and internal warnings should be updated to reflect this
 * [ ] bloom or the API that bloom uses will need to be updated to use a temporary cache that is user-modifiable
 * [ ] The python2 and python3 variants cannot be installed from debs side-by-side; there will be file conflicts in /etc/ and /var/cache/rosdep
 * [ ] The cache may not be readable by python2 and 3. This has not been tested. Since this moves the cache to a new location, this is a good opportunity to try to make the cache cross-version compatible.


wjwwood:
@trainman419 I know we haven't been able to get this merged (it is a big change after all), but any reason you decided to close this just now?

trainman419:
It's been over a year with no constructive input on this, and I'm no longer interested or motivated to work on it.

Issue #339, opened 2014-08-05
rosdep install fails on osx indigo


jmtatsch:
```
~/ros_catkin_ws » rosdep install --from-paths src --ignore-src --rosdistro indigo -y
ERROR: Rosdep experienced an internal error.
Please go to the rosdep page [1] and file a bug report with the message below.
[1] : http://www.ros.org/wiki/rosdep

rosdep version: 0.10.30

Bad installer [homebrew]: Traceback (most recent call last):
  File "/Library/Python/2.7/site-packages/rosdep2/platforms/osx.py", line 182, in is_installed
    pkg_info = json.loads(std_out)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/__init__.py", line 338, in loads
    return _default_decoder.decode(s)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/decoder.py", line 365, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/decoder.py", line 383, in raw_decode
    raise ValueError("No JSON object could be decoded")
ValueError: No JSON object could be decoded
```

jmtatsch:
Now the same thing with jade, repairing the taps did not help.

rosdep install --from-paths src --ignore-src --rosdistro jade -y                                                                                       

ERROR: Rosdep experienced an internal error.
Please go to the rosdep page [1] and file a bug report with the message below.
[1] : http://www.ros.org/wiki/rosdep

rosdep version: 0.11.2

Bad installer [homebrew]: Error while parsing brew info for 'pcl'
 * Output of `brew info pcl --json=v1`:
 
 * Error while parsing:
 Traceback (most recent call last):
  File "/usr/local/lib/python2.7/site-packages/rosdep2/platforms/osx.py", line 184, in is_installed
    pkg_info = json.loads(std_out)
  File "/usr/local/Cellar/python/2.7.10/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/__init__.py", line 338, in loads
    return _default_decoder.decode(s)
  File "/usr/local/Cellar/python/2.7.10/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/decoder.py", line 366, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
[...]

PR #341, opened 2014-08-08
Improve error handling for homebrew installer.

NikolausDemmel:
- Detect when `brew info` fails and provide more meaningful error message.
- Print warnings in situations where rosdep will likely fail to install
  the formula correctly (formula installed but not linked; formula installed
  but with wrong options)

This is meant to address situations like this: https://github.com/ros-infrastructure/rosdep/issues/339#issuecomment-51382416

### DIFF
--- a/src/rosdep2/platforms/osx.py
+++ b/src/rosdep2/platforms/osx.py
@@ -177,6 +177,12 @@ def brew_detect(resolved, exec_fn_err=None, warnings=False):
     :param resolved: List of HomebrewResolution objects
     :returns: Filtered list of HomebrewResolution objects
     """
+    # TODO: Don't raise RosdepInternalError when calls to `brew` fail.
+    #       Instead use some other exception to inform the user what
+    #       happened, but without the request to file a rosdep bugreport.
+    # TODO: It would be better to not print the errors about failed formula
+    #       installations here, but instead somehow pass them up the call
+    #       chain and properly format and print them in one place.
     if exec_fn_err is None:
         exec_fn_err = lambda cmd: read_stdout(cmd, capture_stderr=True, return_exitcode=True)
     std_out, std_err, exitcode = exec_fn_err(['brew', 'list'])


wjwwood:
Despite being almost three years old (🎂?), it seems like a useful change. There are still conflicts and @tfoote's suggestions (which @NikolausDemmel seemed to agree with I think?) were never addressed.

@NikolausDemmel I don't suppose you have any time to resurrect this?

NikolausDemmel:
> Despite being almost three years old (:birthday:?), it seems like a useful change.

:tada:  :grinning: 

Yeah, looks like with rebasing and just slightly changing the output it should be ok. While not perfect, it should be already better than currently, as it seems questions about why rosdep failed in these situations continue to pop up.

> Yeah,  changing the printouts would work. Maybe something like: "ERROR  Details:..." since its additional info for the previously announced  errors.

The only reservation I have here is that in this case the `ERROR Details:` would come before the lines

```
System dependencies have not been satisified:
homebrew    boost --with-python
homebrew    gazebo
homebrew    ogre
```

so it is not referring to a previously announced error.

So maybe just something like the following to keep the implementation simple. It can always be improved later (notice that `Warning: gazebo-1.9.5 already installed, it's just not linked` is the output of brew itself).

```
$ rosdep check --from-paths src --ignore-src --rosdistro indigo --skip-keys python-imaging
ERROR: Formula 'boost' is installed, but used options '' are not
         superset of required options '--with-python'. You might need to manually
         call `brew reinstall boost --with-python` or use `rosdep install ... --reinstall`.
ERROR: Formula 'gazebo' is installed, but not linked. You might need to
         manually call `brew link gazebo` or use `rosdep install ... --reinstall`
ERROR: Formula 'ogre' is installed, but not linked. You might need to
         manually call `brew link ogre` or use `rosdep install ... --reinstall`
System dependencies have not been satisified:
homebrew    boost --with-python
homebrew    gazebo
homebrew    ogre
```

```
$ rosdep install --from-paths src --ignore-src --rosdistro indigo --skip-keys python-imaging
executing command [brew install gazebo]
Warning: gazebo-1.9.5 already installed, it's just not linked
ERROR: Formula 'boost' is installed, but used options '' are not
         superset of required options '--with-python'. You might need to manually
         call `brew reinstall boost --with-python` or use `rosdep install ... --reinstall`.
ERROR: Formula 'gazebo' is installed, but not linked. You might need to
         manually call `brew link gazebo` or use `rosdep install ... --reinstall`
ERROR: Formula 'ogre' is installed, but not linked. You might need to
         manually call `brew link ogre` or use `rosdep install ... --reinstall`
  homebrew: Failed to detect successful installation of [boost --with-python]
  homebrew: Failed to detect successful installation of [gazebo]
  homebrew: Failed to detect successful installation of [ogre]
```

Does that make sense @tfoote ?

I'm not developing on OS X these days, but I'll see what I can do.

NikolausDemmel:
I rebased and updated the PR, but it is untested as of yet.

src/rosdep2/platforms/freebsd.py
```diff
@@ -66,7 +66,7 @@ def pkg_info_detect_single(p):
     pop = subprocess.Popen("/usr/sbin/pkg_info -qE " + portname, shell=True)
     return os.waitpid(pop.pid, 0)[1] == 0 # pkg_info -E returns 0 if pkg installed, 1 if not
 
-def pkg_info_detect(packages):
+def pkg_info_detect(packages, warnings=False):
     return [p for p in packages if pkg_info_detect_single(p)]
 
 class PkgAddInstaller(Installer):
```
NikolausDemmel: Unrelated to this PR: Is it possible that this is outdated? I think it has to derive from `PackageManagerInstaller`, since it even calls the superclass constructor passing the detect function, with `Installer` doesn't have. I guess there are no freebsd rosdep users then, since AFAICT this cannot work.

src/rosdep2/platforms/osx.py
```diff
@@ -179,12 +181,24 @@ def is_installed(r):
...
-        std_out = exec_fn(['brew', 'info', r.package, '--json=v1'])
+        output = exec_fn_err(['brew', 'info', r.package, '--json=v1'])
+        std_out, std_err, exitcode = output
+        if exitcode != 0:
+            raise RosdepInternalError(None, """
+Formula '{0}' is installed, but `brew info` failed with exit code '{1}'.
+Likely the formula script is missing or broken. You can try `brew
+update` and `brew tap --repair` as well as checking the formula script
+for syntax errors (`brew info {0}`). Captured `stderr` output:
+{2}""".format(r.package,  exitcode, std_err))
```
NikolausDemmel: I added that as a TODO note in the source.

PR #348, opened 2014-09-27
Updates to RPM and DNF Package Manager

cottsay:
First commit makes `rosdep` try to use the python API for RPM when detecting packages. If it can't import it (which should never happen), it will fall back to using the current command-based method. The RPM python module is a C extension, and runs quite a bit faster than parsing command line output.

The second commit adds support for the upcoming DNF package manager in Fedora. Eventually we'll want to make this the default installer for F21+, but we'll cross that bridge when we come to it.

### DIFF
--- a/src/rosdep2/platforms/redhat.py
+++ b/src/rosdep2/platforms/redhat.py
@@ -37,10 +37,14 @@ from .source import SOURCE_INSTALLER
 from ..installers import PackageManagerInstaller, TYPE_CODENAME
 from ..shell_utils import read_stdout
 
+# dnf package manager key
+DNF_INSTALLER='dnf'
+
 # yum package manager key
 YUM_INSTALLER='yum'
 
 def register_installers(context):
+    context.set_installer(DNF_INSTALLER, DnfInstaller())
     context.set_installer(YUM_INSTALLER, YumInstaller())
 
 def register_platforms(context):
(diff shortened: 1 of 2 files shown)


georgy7:
@cottsay How to use DnfInstaller with `rosdep` now? Is there a flag, a configuration file or a system variable?

On Fedora 22.

```
rosdep install --from-paths src --ignore-src --rosdistro jade -y
```

Told me, `--skip-broken` command not found (because of `Yum command has been deprecated, redirecting to '/usr/bin/dnf`).

cottsay:
@georgy7 The problem is that we can't just "enable" DNF for all of Fedora. We'll need to add conditional code to use DNF only when the Fedora version is newer than or equal to 22, otherwise we aren't guaranteed that DNF is installed.

I'm not sure what problem you're talking about with `--skip-broken`, but the deprecation was non-fatal last I checked.

Long story short, I'm aware of the issue, but we haven't found a way to cleanly solve it yet. `rosdep` just isn't structured in such a way that it is a straightforward fix...

georgy7:
@cottsay 
I mean, when I run `rosdep install`, Fedora redirects yum command to dnf command, that has not `--skip-broken` flag. When I read 00600b3 patch, I saw that it already have dnf commands without `--skip-broken` flag in `DnfInstaller` class. But by default, rosdep uses `YumInstaller` class. Is there a way to forced switch to `DnfInstaller` at least only once, for the single command?

cottsay:
Okay, I can confirm the issue that @georgy7 is describing. Evidently the backwards compatibility with `yum` isn't as robust as we'd hoped, and even though `--skip-broken` is now the default behavior in `dnf`, supplying the flag causes `dnf` to fail, even when invoked via the `yum` backwards compatibility.

In the meantime, you could install the actual `yum` package on the machine with `sudo dnf install yum`. I'll try to make time this weekend to devise a solution for enabling `dnf` as the default for `f22`+.

georgy7:
Thanks!
I will wait the fix in rosdep (instead of installing yum, etc), because I afraid to break my Fedora installation, and it is not so urgent. Also, thus I will be able to confirm, the issue is fixed.

PR #361, opened 2014-11-25
raise CachePermissionError on rosdep install, too

k-okada:


wjwwood:
I'm going to close this due to inactivity. It can be reopened if needed.

Issue #368, opened 2015-01-22
rosdep cannot detect installation of [argparse]


jmtatsch:
When I run `rosdep install --from-paths src --ignore-src --rosdistro indigo -y` I get:

```
executing command [sudo pip install -U argparse]
The directory '/Users/tatsch/Library/Logs/pip' or its parent directory is not owned by the current user and the debug log has been disabled. Please check the permissions and owner of that directory. If executing pip with sudo, you may want the -H flag.
The directory '/Users/tatsch/Library/Caches/pip/http' or its parent directory is not owned by the current user and the cache has been disabled. Please check the permissions and owner of that directory. If executing pip with sudo, you may want the -H flag.
The directory '/Users/tatsch/Library/Caches/pip/http' or its parent directory is not owned by the current user and the cache has been disabled. Please check the permissions and owner of that directory. If executing pip with sudo, you may want the -H flag.
Requirement already up-to-date: argparse in /Library/Python/2.7/site-packages
ERROR: the following rosdeps failed to install
  pip: Failed to detect successful installation of [argparse]
```

tfoote:
What is your platform? 

Can you run the command it's trying to run manually?

Can you make a one line command which does not use the source directory or provide us with a rosinstall for your source directory for us to reproduce?

jmtatsch:
OSX Yosemite / ROS Indigo

If I run the command manually:

````
 sudo pip install -U argparse
Password:
The directory '/Users/tatsch/Library/Logs/pip' or its parent directory is not owned by the current user and the debug log has been disabled. Please check the permissions and owner of that directory. If executing pip with sudo, you may want the -H flag.
The directory '/Users/tatsch/Library/Caches/pip/http' or its parent directory is not owned by the current user and the cache has been disabled. Please check the permissions and owner of that directory. If executing pip with sudo, you may want the -H flag.
The directory '/Users/tatsch/Library/Caches/pip/http' or its parent directory is not owned by the current user and the cache has been disabled. Please check the permissions and owner of that directory. If executing pip with sudo, you may want the -H flag.
Requirement already up-to-date: argparse in /usr/local/lib/python2.7/site-packages
````

Rosinstall is here:

https://gist.github.com/jmtatsch/a4ff114be8378e5f32c4

tfoote:
Looking closer the first 3 lines appear to be warnings then it says the requirement is already up to date. But it appears to be returning a non-zero return code. 

[...]

PR #369, opened 2015-01-22
use sudo -H to make pip happy

wjwwood:
See: #368


tfoote:
This looks reasonable as a default, though there are a bunch of unit tests failing.

Issue #10, opened 2012-06-23
ROS_OS_OVERRIDE is not parsed right


davidwe:
I am getting different results in this two cases, where i think the programm should act in the same way:

case 1: "rosdep install -a --os="debian:squeeze"
case 2: "rosdep install -a" with the ROS_OS_OVERRIDE environment variable set to "debian:squeeze"

I suppose, in case 1 rosdep is working correctly, while in case 2 i get the following error for a example:
package: No definiton of [boost] for OS version[]

Issue #366, opened 2014-12-26
Address overlap of CentOS and RHEL


wjwwood:
See: https://github.com/ros-infrastructure/rospkg/pull/76

Issue #272, opened 2013-09-18
rosdep could use upstream LSB data


gavanderhoorn:
I have a machine running [Elementary OS - Luna](http://elementaryos.org/). This is an Ubuntu Precise derivative, and `lsb_release -a` returns:

```
Distributor ID:  elementary OS
Description:     elementary OS Luna
Release:         0.2
Codename:        luna
```

Not having any rules for that, `rosdep` never finds anything unless I set the `ROS_OS_OVERRIDE` environment variable (but this has issues too, see #51 fi).

`lsb_release -ua` however, returns the following:

```
Distributor ID:  Ubuntu
Description:     Ubuntu 12.04 LTS
Release:         12.04
Codename:        precise
```

Would it perhaps be an idea for `rosdep` to try to parse upstream LSB data if it can't find a match using the normal invocation of `lsb_release`?

A command line option might be required to be able to disable (or enable) this behaviour.

Issue #159, opened 2013-01-07
Add Linaro as an alias for Ubuntu to satisfy rosdeps correctly  (ros ticket #3534)


tfoote:
As noted in this question (http://answers.ros.org/question/1191/installing-camera-drivers-on-arm), rosdep does not know how to handle the OS identifier "Linaro" when in fact Linaro-based systems are effectively Ubuntu systems.
The impact of this change is currently small (ARM systems) but the fix seems to be adding debian.Linaro to the os_list in the Rosdep class init (tools/rosdep/src/rosdep/core.py) and then add this in to the debian.py file as has been done with Mint.  Versioning on this variant will change in future but for now, it'll be the same as Ubuntu.

trac data:
- Owner: **tfoote**
- Reporter: **ashcharles**
- Reported at: **Sun Jun 12 22:27:53 2011**
- URL: http://code.ros.org/trac/ros/ticket/3534

PR #370, opened 2015-01-23
Linaro support

tfoote:
Add an alias for linaro to ubuntu. It will only trigger if the OS is not overridden. (This does not resolve: #10 it only works with --os overrides)

This looks like a potential solution for #272 and similar issues #366

### DIFF
--- a/src/rosdep2/platforms/debian.py
+++ b/src/rosdep2/platforms/debian.py
@@ -28,7 +28,10 @@
 
 # Author Tully Foote, Ken Conley
 
-from rospkg.os_detect import OS_DEBIAN, OS_UBUNTU, OsDetect
+from __future__ import print_function
+import sys
+
+from rospkg.os_detect import OS_DEBIAN, OS_LINARO, OS_UBUNTU, OsDetect
 
 from .pip import PIP_INSTALLER
 from .gem import GEM_INSTALLER
@@ -44,6 +47,7 @@ def register_installers(context):
 
 def register_platforms(context):
     register_debian(context)
+    register_linaro(context)
     register_ubuntu(context)
     
 def register_debian(context):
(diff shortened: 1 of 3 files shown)


tfoote:
Travis will fail until rospkg is released.

wjwwood:
You need to add the version dependency to the `stdeb.cfg` as well: https://github.com/ros-infrastructure/rosdep/blob/master/stdeb.cfg#L2-L3

tfoote:
stdeb.cfg updated.

Also this is a proper fix for #159 instead of documentation.

src/rosdep2/platforms/debian.py
```diff
@@ -54,6 +55,13 @@ def register_debian(context):
     context.set_default_os_installer_key(OS_DEBIAN, APT_INSTALLER)
     context.set_os_version_type(OS_DEBIAN, OsDetect.get_codename)
     
+def register_linaro(context):
+    # Linaro is an alias for Ubuntu. If linaro is detected and it's not set as an override force ubuntu.
+    (os_name, os_version) = context.get_os_name_and_version()
+    if os_name == OS_LINARO and not context.os_override:
+        print("rosdep detected OS: [%s] aliasing it to: [%s]" % (OS_LINARO, OS_UBUNTU))
```
wjwwood: Should it always print here? Maybe it should, I'm not sure though.

wjwwood:
Otherwise, looks good, +1.

src/rosdep2/platforms/debian.py
tfoote: I was going back and forth. I put it in as a debugging statement, but I thought that it was important to let the user know it's using a different set of rules. Otherwise they would have no introspection into the fact that it's cross referencing. Maybe it should go to stderr if people are scraping the console output of things like resolve?

wjwwood: If it always prints, I would at least put it to stderr. I agree that it should be visible though.

tfoote:
Moved output to stderr

gavanderhoorn:
@tfoote: iiuc, this only works for `rosdep` related problems with unrecognised OS right? It's not a general solution to detecting derivatives of already supported OS?

(or is that what *"This does not resolve: #10 it only works with --os overrides"* actually says?)

tfoote:
This is not a generic solution for unrecognized OSs, but is a pattern which derivative OSs can declare the cross reference to another tag. 

My comment re #10 is that if you set ROS_OS_OVERRIDE this code path will still override if you set `ubuntu:trusty` whereas is you use --os option it will accept your override.

tfoote:
@wjwwood can you review and merge and then we can release? This is blocking me making the armhf stuff mainline. let me know and I'm happy to run the release myself.

tfoote:
The tests are passing now that rospkg has been released.

wjwwood:
LGTM, I was wanting to get the fix/workaround for detection of `argparse` from `pip` working before releasing, but if you need it now, you can go ahead and release and I'll do another release soon with the `argparse` fix.

PR #374, opened 2015-02-13
add options to printout only apt/pip instllable packages, for db command

k-okada:


wjwwood:
@k-okada you're not giving me much to go on here, I'd like to know:

- Why do you need this?
- What is the intended behavior of the ?
- What's an example of using this in a practical situation?
- Is this related to some other issue/pull request?

General rule of thumb, if you open a pull request with no description (only a subject), then it better be *really* obvious what is being addressed, like a typo or something. If you're adding a feature, then you need to justify it.

src/rosdep2/main.py
```diff
@@ -296,6 +296,10 @@ def _rosdep_main(args):
                            "locally ignore a rosdep key is creating a local rosdep rule "
                            "with an empty list of packages (include it in "
                            "/etc/ros/rosdep/sources.list.d/ before the defaults).")
+    parser.add_option("--install-method",
```
wjwwood: I think this needs a verb, a good way to tell is to read out loud the finial command name, "rodep db --install-method", doesn't really say what `db` is doing with the provided install method. What about `--filter-for-install-method INSTALL_METHOD`? Also should it be `install-method` or `installer`?

src/rosdep2/main.py
```diff
@@ -296,6 +296,10 @@ def _rosdep_main(args):
...
+                      dest='inst_key', default=None,
```
wjwwood: Why use a different destination?

src/rosdep2/main.py
```diff
@@ -296,6 +296,10 @@ def _rosdep_main(args):
...
+                      help="Affects the 'db' verbs. db command print out packages only "
```
wjwwood: It should be `Affects the 'db' verb.`, the `verb` shouldn't be plural.

I would reword the second sentence like this: `filters the output of the 'db' command by packages installed with the provided install method; can be one of <installers>`, then I would fill `<installers>` with the actual list of installers. You should be able to build it. Failing that I would work the last bit `... install method; can be any installer, e.g. apt, pip, etc.`.

src/rosdep2/main.py
```diff
@@ -678,6 +682,8 @@ def command_db(options):
         try:
             d = view.lookup(rosdep_name)
             inst_key, rule = d.get_rule_for_platform(os_name, os_version, installer_keys, default_key)
+            if options.inst_key and inst_key != options.inst_key :
```
wjwwood: No extra space before the `:`.

k-okada:
this PR is for https://github.com/ros-infrastructure/jenkins_scripts/pull/77, which we want to create doc pages even inf they have pip package on build/run depend

k-okada:
also updated source code by you comments

NikolausDemmel:
Sorry to be nitpicky, but I also have an opinion about the wording and the argument name.

@k-okada: Please consider [my PR](https://github.com/k-okada/rosdep/pull/1) to your branch (this will update your PR to rosdep master). I believe this is in line with @wjwwood 's comments.

This changes:
- Allow multiple installers to be passed; make it work similar to `--skip-keys`.
- Change option name. The term "install method" is not used so far in rosdep. Instead, we call it "installer" to avoid confusion.

wjwwood:
lgtm

PR #386, opened 2015-07-04
Added threaded building flag to emerge command on Gentoo.

allenh1:
This is a very small change, but threading emerge is a very excellent command. This makes building the large amount of packages that ros pulls in build MUCH faster.


tfoote:
Unlimited parallelization can be a problem. We've seen this make. To add this we should probably at least use a --load-average limiter. It's not optimal but hopefully will at least prevent most overloading issues.

allenh1:
I suppose the next question is what the load average should be... Do you think we could use the number output by nproc?  Something like `--jobs --load-average 'nproc'`, substituting ' for `.

allenh1:
The other complaint I have about he way rosdep uses emerge is the fact that often the USE flags are not satisfied. It is an easy fix, but another tag...  `--autounmask-write`, but then it is necessary to perform an `etc-update`.

tfoote:
yeah, a load-average limit of nproc is much better. Though you can't use backticks. You should do it in python [this thread](https://stackoverflow.com/questions/1006289/how-to-find-out-the-number-of-cpus-using-python) has several ways. You should find one which adds the fewest dependencies. 

The USE flags should be considered in another thread.

allenh1:
Sounds good. I updated the code to include the appropriate flags.

tfoote:
lgtm @wjwwood thoughts?

wjwwood:
Find by me, I don't have a Gentoo machine to test this at all.

src/rosdep2/platforms/gentoo.py
```diff
@@ -114,6 +115,11 @@ def get_install_command(self, resolved, interactive=True, reinstall=False, quiet
         if not atoms:
             return []
 
+        # The emerge command can be threaded.
+        cmd.append('--jobs')
+        # The next command prevents unlimited parallelization
+        cmd.append('--load-average {0}'.format(multiprocessing.cpu_count()))
```
wjwwood: Before I merge this, please handle the case where `multiprocess.cpu_count()` can raise `NotImplementedError`:

https://docs.python.org/2/library/multiprocessing.html#multiprocessing.cpu_count

allenh1:
I just did a simple try block that adds the flags if it can count processors, and doesn't otherwise. It just prints a warning that the build will not be threaded. Let me know if this should be handled in a different manner!

allenh1:
I'm going to tag one more thing onto this commit. There's a Gentoo flavor that is much better about supporting hardware. It behaves in the exact same manner as Gentoo (uses portage and everything). The install is just set up for getting you in a desktop faster. Great simplifications of things. It's called Funtoo. The only reason I have to add something comes from the fact that rosdep checks the `/etc/gentoo-release` file (which exists on Funtoo installs, but has slightly different contents). I installed ROS on Funtoo by replacing the `/etc/gentoo-release` file's contents with a copy of one from one of my Gentoo installs.

allenh1:
After digging through source files, the suggestion I just made goes in `rospkg` (specifically, `os_detect.py`). This should be ready for merge. @tfoote @wjwwood, thoughts?

tfoote:
I followed up at https://github.com/ros-infrastructure/rospkg/pull/82#issuecomment-120876864

This looks reasonable. We'll try to merge and release when we get some free time.

wjwwood:
Several of the tests will continue to fail until the change is made in rospkg (I guess). Let's not forget to re-trigger the tests here once that's done.

tfoote:
I merged the changes in rospkg but it needs to be released. And this will need a version dependency on the new version or higher.

allenh1:
Sounds good! Thanks!

wjwwood:
@tfoote is the required change released into `rospkg` yet?

tfoote:
Yes merged and released

wjwwood:
Ok, then I restarted the Travis-CI job.

allenh1:
Hm... @tfoote  Any idea as to why this isn't building?

NikolausDemmel:
Some how tully's comment doesn't link here, so here it is: https://github.com/ros-infrastructure/rosdep/commit/a6c5aa8b464ed1f71a9dc0e9133d610271dd45d9#commitcomment-14866900

tfoote:
Thanks @NikolausDemmel I had assumed the comment would make it here. @allenh1 that comment is why the CI is failing.

allenh1:
WOOT. It passed. I'll run a quick rebase on the commits, and it should be ready.

tfoote:
It looks good. A rebase would be great then we can merge.

allenh1:
Ok - squashed.

NikolausDemmel:
+1

src/rosdep2/platforms/gentoo.py
```diff
@@ -54,10 +58,23 @@ def register_installers(context):
...
     context.add_os_installer_key(OS_GENTOO, PORTAGE_INSTALLER)
     context.add_os_installer_key(OS_GENTOO, SOURCE_INSTALLER)
     context.set_default_os_installer_key(OS_GENTOO, PORTAGE_INSTALLER)
 
+def register_funtoo(context):
+    # Funtoo is an alias for Gentoo.
+    (os_name, os_version) = context.get_os_name_and_version()
+    if os_name == OS_FUNTOO and not context.override:
+        print("rosdep detected OS: [%s], aliasing it to [%s]" % (OS_FUNTOO, OS_GENTOO), file=sys.stderr)
+        context.add_os_installer_key(OS_GENTOO, PORTAGE_INSTALLER)
```
tfoote: Is this not just duplicating the above gentoo block? These exact installer keys were already registered above.

allenh1: Do you want me to call the gentoo context for the below two lines then?

NikolausDemmel: You probably want to set the OS override like here: https://github.com/ros-infrastructure/rosdep/blob/91816c59a90f22f0044ef0c4843138d8f71d068e/src/rosdep2/platforms/debian.py#L62-L67

allenh1:
Here's what I'm getting upon testing. It was unable to alias the OS, but it definitely detected that it was running on funtoo.

```
allenh1@tuxbox ~/ros_catkin_ws/src/rosdep/scripts $ ./rosdep install --from-paths ~/catkin_ws/src --ignore-src --rosdistro jade -y
ERROR: the following packages/stacks could not have their rosdep keys resolved
to system dependencies:
rosconsole: Unsupported OS [funtoo]
catkin: Unsupported OS [funtoo]
image_view: Unsupported OS [funtoo]
urdf: Unsupported OS [funtoo]
rosgraph: Unsupported OS [funtoo]
resource_retriever: Unsupported OS [funtoo]
nodelet_topic_tools: Unsupported OS [funtoo]
rqt_graph: Unsupported OS [funtoo]
qt_gui: Unsupported OS [funtoo]
roslib: Unsupported OS [funtoo]
rosbuild: Unsupported OS [funtoo]
rosclean: Unsupported OS [funtoo]
tf: Unsupported OS [funtoo]
convex_decomposition: Unsupported OS [funtoo]
roswtf: Unsupported OS [funtoo]
geometric_shapes: Unsupported OS [funtoo]
rqt_web: Unsupported OS [funtoo]
robot_state_publisher: Unsupported OS [funtoo]
rqt_py_console: Unsupported OS [funtoo]
rqt_topic: Unsupported OS [funtoo]
rqt_top: Unsupported OS [funtoo]
rqt_rviz: Unsupported OS [funtoo]
pluginlib: Unsupported OS [funtoo]
rqt_msg: Unsupported OS [funtoo]
rqt_service_caller: Unsupported OS [funtoo]
rqt_pose_view: Unsupported OS [funtoo]
image_geometry: Unsupported OS [funtoo]
pcl_conversions: Unsupported OS [funtoo]
image_rotate: Unsupported OS [funtoo]
rqt_console: Unsupported OS [funtoo]
depth_image_proc: Unsupported OS [funtoo]
python_qt_binding: Unsupported OS [funtoo]
laser_geometry: Unsupported OS [funtoo]
rviz: Unsupported OS [funtoo]
rqt_bag: Unsupported OS [funtoo]
qt_gui_py_common: Unsupported OS [funtoo]
eigen_conversions: Unsupported OS [funtoo]
diagnostic_common_diagnostics: Unsupported OS [funtoo]
camera_calibration_parsers: Unsupported OS [funtoo]
laser_filters: Unsupported OS [funtoo]
camera_calibration: Unsupported OS [funtoo]
rostime: Unsupported OS [funtoo]
rospy: Unsupported OS [funtoo]
rosunit: Unsupported OS [funtoo]
python_orocos_kdl: Unsupported OS [funtoo]
stage: Unsupported OS [funtoo]
bondcpp: Unsupported OS [funtoo]
rosconsole_bridge: Unsupported OS [funtoo]
camera_info_manager: Unsupported OS [funtoo]
roslz4: Unsupported OS [funtoo]
rqt_bag_plugins: Unsupported OS [funtoo]
rosmsg: Unsupported OS [funtoo]
turtlesim: Unsupported OS [funtoo]
rqt_robot_monitor: Unsupported OS [funtoo]
rosparam: Unsupported OS [funtoo]
rospack: Unsupported OS [funtoo]
tf2: Unsupported OS [funtoo]
rqt_runtime_monitor: Unsupported OS [funtoo]
rqt_dep: Unsupported OS [funtoo]
rqt_plot: Unsupported OS [funtoo]
rostest: Unsupported OS [funtoo]
cpp_common: Unsupported OS [funtoo]
nodelet: Unsupported OS [funtoo]
stage_ros: Unsupported OS [funtoo]
rqt_robot_steering: Unsupported OS [funtoo]
roslaunch: Unsupported OS [funtoo]
genpy: Unsupported OS [funtoo]
rqt_shell: Unsupported OS [funtoo]
roscreate: Unsupported OS [funtoo]
qt_dotgraph: Unsupported OS [funtoo]
rqt_publisher: Unsupported OS [funtoo]
cv_bridge: Unsupported OS [funtoo]
gazebo_ros: Unsupported OS [funtoo]
rosbag_storage: Unsupported OS [funtoo]
joint_state_publisher: Unsupported OS [funtoo]
octomap: Unsupported OS [funtoo]
stereo_image_proc: Unsupported OS [funtoo]
bondpy: Unsupported OS [funtoo]
tf2_kdl: Unsupported OS [funtoo]
class_loader: Unsupported OS [funtoo]
collada_urdf: Unsupported OS [funtoo]
pcl_ros: Unsupported OS [funtoo]
rosmake: Unsupported OS [funtoo]
image_proc: Unsupported OS [funtoo]
roscpp: Unsupported OS [funtoo]
collada_parser: Unsupported OS [funtoo]
actionlib: Unsupported OS [funtoo]
qt_gui_cpp: Unsupported OS [funtoo]
eigen_stl_containers: Unsupported OS [funtoo]
pr2_description: Unsupported OS [funtoo]
roslisp: Unsupported OS [funtoo]
message_filters: Unsupported OS [funtoo]
random_numbers: Unsupported OS [funtoo]
gazebo_plugins: Unsupported OS [funtoo]
tf_conversions: Unsupported OS [funtoo]
dynamic_reconfigure: Unsupported OS [funtoo]
rqt_tf_tree: Unsupported OS [funtoo]
rqt_logger_level: Unsupported OS [funtoo]
theora_image_transport: Unsupported OS [funtoo]
urdf_parser_plugin: Unsupported OS [funtoo]
orocos_kdl: Unsupported OS [funtoo]
rosbag: Unsupported OS [funtoo]
```

allenh1:
Wait -- I just have no idea what I'm doing with testing apparently. Apologies. Please disregard the above.

allenh1:
So, upon testing, I encountered the following: 
```
ERROR: OS override must be colon-separated OS_NAME:OS_VERSION, e.g. ubuntu:maverick
```

NikolausDemmel:
Are you testing by manually setting a override on the command line? Are the codenames returned by OSDetect the same for Gentoo and Funtoo?

allenh1:
roapkg definitely identifies funtoo as funtoo. The error was thrown by just running `nosetests` in the test directory for rosdep.

NikolausDemmel:
Could you post the full command and resulting output to gist and link it here?

Additionally, what is the output of `python -c "import rospkg.os_detect; print(rospkg.os_detect.OsDetect().detect_os())"` on funtoo?

allenh1:
I'm about to post the gist. Here's the output of the python command. It looks alright, as far as I can tell.

```
('funtoo', u'2.20', '')
```

allenh1:
https://gist.github.com/allenh1/5f2fae4b6cd10ebe239b

src/rosdep2/platforms/gentoo.py
```diff
@@ -54,10 +58,21 @@ def register_installers(context):
...
     context.add_os_installer_key(OS_GENTOO, PORTAGE_INSTALLER)
     context.add_os_installer_key(OS_GENTOO, SOURCE_INSTALLER)
     context.set_default_os_installer_key(OS_GENTOO, PORTAGE_INSTALLER)
 
+def register_funtoo(context):
+    # Funtoo is an alias for Gentoo.
+    (os_name, os_version) = context.get_os_name_and_version()
+    if os_name == OS_FUNTOO and not context.override:
+        print("rosdep detected OS: [%s], aliasing it to [%s]" % (OS_FUNTOO, OS_GENTOO), file=sys.stderr)
+        context.set_os_override(OS_GENTOO, context.os_detect.get_codename())
```
NikolausDemmel: Ah so it seems the codename is not defined for gentoo/funtoo and always returns the empty string `''`. Try `get_version()` instead of `get_codename()` here for the override.

allenh1:
It has come to my attention that the testcases need to change slightly because of the things added to the emerge command. I am working on that at the moment.

allenh1:
Ok! Everything should be good now. I updated the gist in the above link, and, as you can see, the only issue is the fact that it prints the fact that it is aliasing to gentoo. @tfoote I think it's ready!

NikolausDemmel:
Hm... we may want to adapt the unit tests such that they don't fail. It looks like they must be failing in a similar fashion for Linaro and Elemetary. Alternatively we could remove the override warning to stderr.

allenh1:
I'll update the main test, if needed.

NikolausDemmel:
@tfoote, what do you think?

I guess the main test could allow that specific warning on Funtoo (and Linaro and Elementary as well).

tfoote:
It would be great to have the unit tests know about this and not fail on the platforms where we're aliasing.

allenh1:
You got it!y basic strategy is to capture stderr into a variable and replace all instances of the platform substring with a null string. Is this good with everyone?

tfoote:
That seems reasonable.

allenh1:
@tfoote I think this one is ready for merge -- do you remember if there was any reason we didn't merge it?

src/rosdep2/platforms/gentoo.py
```diff
@@ -54,10 +58,21 @@ def register_installers(context):
...
 def register_platforms(context):
+    register_funtoo(context)
+    register_gentoo(context)
+
+def register_gentoo(context):
     context.add_os_installer_key(OS_GENTOO, PORTAGE_INSTALLER)
     context.add_os_installer_key(OS_GENTOO, SOURCE_INSTALLER)
     context.set_default_os_installer_key(OS_GENTOO, PORTAGE_INSTALLER)
 
+def register_funtoo(context):
```
tfoote: Is the Funtoo change unrelated and just snuck in here?

wjwwood:
This still needs a rebase perhaps. For one reason or another CI is still failing.

src/rosdep2/platforms/gentoo.py
allenh1: I think I was trying to kill two issues with one PR here...

Not even sure where this code is, anymore (has it _really_ been two years??).

PR #392, opened 2015-07-28
Update command documentation with latest options

cmansley:
Add the rest of the options from rosdep -h


wjwwood:
@cmansley thanks for updating these, we've definitely been negligent since adding new options.

PR #393, opened 2015-07-29
commands.rst: Path to sources was incorrect.

aymanim:
The actual sources list is at /etc/ros/rosdep/sources.list.d and not /etc/ros/sources.list.d.


dirk-thomas:
Thanks.

Issue #382, opened 2015-04-29
Make `fix-permissions` work when installation is not initialized


NikolausDemmel:
I just wanted to initialize rosdep on a new machine.

I called `sudo rosdep init` and then `sudo rosdep update`, then I realized that I should have run the second command without `sudo`, so I aborted before it completed. Now `rosdep update` shows the following error

```
reading in sources list data from /etc/ros/rosdep/sources.list.d
Failed to write cache file: [Errno 13] Permission denied: '/home/den2pal/.ros/rosdep/sources.cache/a4b3c8956ab7a3e6d293ba33b57fac42799e69e6.pickle.tmp.U2Ww_A'
Try running 'sudo rosdep fix-permissions'
```

and `sudo rosdep fix-permissions` shows 

```
ERROR: your rosdep installation has not been initialized yet.  Please run:
        
    rosdep update
```

It should be easily possible to change `fix-permissions` to work also in this uninitialized state (I don't have time to implement right away though).

wjwwood:
Thanks for the report, I also don't have time at the moment, I'll throw it in untargeted.

PR #398, opened 2015-08-13
Make fix-permissions work in an uninitialized state

spaghetti-:
Fixes https://github.com/ros-infrastructure/rosdep/issues/382

Tested on Ubuntu/OSX


NikolausDemmel:
Thanks for the patch.

Not sure we need to setup the proxy for fix-permissions.

The patch looks like it solves the problem. Could you possibly rewrite it to neither setup the proxy, nor call `check_for_sources_list_init(options.sources_cache_dir)` to avoid confusion later?

spaghetti-:
@NikolausDemmel sure, done.

NikolausDemmel:
lgtm :+1:

tfoote:
Thanks for the patch and iteration.

Issue #378, opened 2015-04-03
rosdep permissions


mdushkoff:
So the permissions were messed up with rosdep and it suggested to "Try running 'sudo rosdep fix-permissions'" to fix this, but when I did so the following error was produced on a machine running Ubuntu 14.04:

```
ERROR: Rosdep experienced an error: 'getgrgid(): gid not found: 15318'
Please go to the rosdep page [1] and file a bug report with the stack trace below.
[1] : http://www.ros.org/wiki/rosdep

rosdep version: 0.11.2

Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/rosdep2/main.py", line 136, in rosdep_main
    exit_code = _rosdep_main(args)
  File "/usr/lib/python2.7/dist-packages/rosdep2/main.py", line 343, in _rosdep_main
    return _no_args_handler(command, parser, options, args)
  File "/usr/lib/python2.7/dist-packages/rosdep2/main.py", line 351, in _no_args_handler
    return command_handlers[command](options)
  File "/usr/lib/python2.7/dist-packages/rosdep2/main.py", line 769, in command_fix_permissions
    group_name = grp.getgrgid(gid).gr_name
KeyError: 'getgrgid(): gid not found: 15318'
```

wjwwood:
Thanks for the report, I'll have a look into the issue asap.

@Mdu7078 As a work around you can probably try to remove the folder `~/.ros/rosdep` and then do `rosdep update`.

@NikolausDemmel I believe you contributed this verb for rosdep, does anything jump out at you here that might help me narrow down the issue? (I haven't looked into it at all yet)

tfoote:
One things that sticks out is that the group id is very high. Below 1000 is reserved and typically new groups start being numbered at 1000 and incrementing upward.

NikolausDemmel:
Could this be to do with LDAP groups? Are you in a corporate net (or similar) with LDAP authentication?

NikolausDemmel:
@Mdu7078 Could you please post the output if running `id` on the command line?

Mdu7078:
I am running this on a server that has a very large number of users.

This is the output of `id`:
uid=15318(mdu7078) gid=3128(domain users) groups=3128(domain users)

Mdu7078:
@wjwwood Removing `~/.ros/rosdep` and then runing `rosdep update` did not work as a workaround and instead produces this error:

```
[...]

PR #399, opened 2015-08-18
Check if group name is available, if not fall back to gid

spaghetti-:
See: https://github.com/ros-infrastructure/rosdep/issues/378


wjwwood:
I think this should work, thanks @spaghetti-. It certainly isn't any worse than what's already there.

PR #403, opened 2015-08-26
Use dnf as default installer for Fedora

sorki:
Yum is no more since F22 - there's only wrapper for dnf that doesn't support yum's parameters.


cottsay:
DNF is not necessarily installed on Fedora 21 systems, and Fedora 21 is still supported. I'd rather develop a solution that does the "right thing" on for all of Fedora...

sorki:
Closing in favour of #404

PR #404, opened 2015-08-26
Use DNF for Fedora 22+

sorki:
Closes #403

Signed-off-by: Richard Marko <rmarko@fedoraproject.org>

PR #405, opened 2015-08-28
Fixes for Fedora/DNF

cottsay:
Various fixes for Fedora.

First (and foremost), make DNF the default installer for f22+. This is solved in the same exact manner as the key problem was solved for Fedora 21 (see https://github.com/ros-infrastructure/rosdep/pull/349), so the code should look consistent.

The only API change is that `set_default_os_installer_key` now takes in a method that returns an installer key, instead of just an installer key.

I tested simple resolution, unit tests, and Bloom generation on:
- Fedora 21
- Fedora 22
- Ubuntu Trusty

If @wjwwood could give it a quick once-over from an OSX perspective, that would be great, though I don't anticipate any complications there.

Thanks all,

--scott

Alternate to https://github.com/ros-infrastructure/rosdep/pull/404

Tagging @jpgr87

### DIFF
--- a/src/rosdep2/installers.py
+++ b/src/rosdep2/installers.py
@@ -211,13 +211,12 @@ class InstallerContext(object):
         """
         if not os_key in self.os_installers:
             raise KeyError("unknown OS: %s"%(os_key))
-        if not installer_key in self.os_installers[os_key]:
-            raise KeyError("installer [%s] is not associated with OS [%s]. call add_os_installer_key() first"%(installer_key, os_key))
-
-        # validate, will throw KeyError
-        self.get_installer(installer_key)
+        if not hasattr(installer_key, '__call__'):
+            raise ValueError("version type should be a method")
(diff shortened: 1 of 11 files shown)


wjwwood:
Code looks good, and seems to run for me on OS X. +1

I'll try to get a release out this week (I also need to do wstool and friends).

RobbieTheK:
Any reason why https://apps.fedoraproject.org/packages/gazebo/  is broken?

Service Unavailable
The server is temporarily unable to service your request due to maintenance downtime or capacity problems. Please try again later.

Apache Server at apps.fedoraproject.org Port 443

cottsay:
> Any reason why https://apps.fedoraproject.org/packages/gazebo/ is broken?

That service is unavailable right now, probably due to the ongoing datacenter migration happening at the Fedora Project: https://status.fedoraproject.org/

PR #411, opened 2015-09-15
Fix pip pkg detection way for old pip client

### LINKED ISSUE #410: rosdep does not resolve python dependencies when having old version pip
```
% rosdep check jsk_data
...
resolve_all: resource [jsk_data] requires rosdep keys [catkin, python-freezegun-pip, python-nose]
loading view [*default*] with rospkg loader
resolution: pip [freezegun]
uninstalled: []
resolution: apt [ros-indigo-catkin, python-nose]
uninstalled: []
All system dependencies have been satisified 
```
It says nothing to install. but
```
% pip freeze | grep freeze
% pip show freeze
```
freeze package has not been installed yet.

I think this is because of these lines.
https://github.com/ros-infrastructure/rosdep/blob/master/src/rosdep2/platforms/pip.py#L73-L82
it checkes if the package is installed using `pip show -q {pkg}`, and latest pip returns 1 as error code when the specified package is not yet installed, but the old pip (which is installed via apt on ubuntu trusty) always returns 0. My env is
```
% lsb_release -r
Release:	14.04
% pip -V
pip 1.5.4 from /usr/lib/python2.7/dist-packages (python 2.7)
% which pip
/usr/bin/pip
% dpkg --get-selections | grep python-pip
python-pip					install
```

this seems related to https://github.com/ros-infrastructure/rosdep/pull/371 .
I noticed at: https://github.com/jsk-ros-pkg/jsk_common/pull/1120#partial-pull-merging

wkentaro:
Closes #410
For https://github.com/jsk-ros-pkg/jsk_common/pull/1120

### DIFF
--- a/src/rosdep2/platforms/pip.py
+++ b/src/rosdep2/platforms/pip.py
@@ -77,7 +77,9 @@ def pip_detect(pkgs, exec_fn=None):
     #   https://github.com/pypa/pip/issues/1570#issuecomment-71111030
     if fallback_to_pip_show:
         for pkg in [p for p in pkgs if p not in ret_list]:
-            if subprocess.call(['pip', 'show', '-q', pkg]) == 0:
+            # does not see retcode but stdout for old pip to check if installed
+            output = subprocess.check_output(['pip', 'show', pkg]).strip()
+            if not output:
                 # `pip show` detected it, add it to the list.
                 ret_list.append(pkg)
 


wkentaro:
@wjwwood 
I hope a release for this patch.
Is that possible or should I wait for a while?

wjwwood:
@wkentaro `rosdep` has been on my plate to release for a while now. I'll try to do a release soon.

PR #413, opened 2015-09-25
Correct logic for wrong package detection with 'pip show'

wkentaro:
Fix #411


wkentaro:
@wjwwood 
I'm really sorry about this.
I misunderstood about the function and the logic condition should be opposite.
(if package is installed, the output of `pip show` is true and should be added to `ret_list`)

wjwwood:
That's ok, I'll just make another release. Thanks for the quick detection and fix.

wjwwood:
@wkentaro I just checked https://github.com/ros-infrastructure/rosdep/issues/414 against this patch and it still happens. I'll look into it.

PR #419, opened 2015-09-27
added pip installare to yocto-edison os

alessandro-aglietti:
when os_detect is OS_YOCTO_EDISON (see https://github.com/ros-infrastructure/rospkg/pull/86) is need using only pip to install modules


wjwwood:
Closing with https://github.com/ros-infrastructure/rospkg/pull/86 and due to conflicts.

PR #439, opened 2016-03-18
Default to urllib2 to make urlopen more robust

efernandez:
With urllib (HTTP/1.0) rosdep update intermittently fails:

``` bash
$ for i in $(seq 1 100); do rosdep update; done
```

With urllib2 (HTTP/1.1) it doesn't (at least on my tests):

``` python
$ ipython
In [1]: from rosdep2.main import rosdep_main
In [2]: while True:
    rosdep_main(["update"])
```

See this too (there's probably a better reference):
http://stackoverflow.com/questions/9422231/why-does-urllib-urlopenurl-fail-while-urllib2-urlopenurl-works-what-specifi

### DIFF
--- a/src/rosdep2/sources_list.py
+++ b/src/rosdep2/sources_list.py
@@ -35,11 +35,11 @@ import tempfile
 import yaml
 import hashlib
 try:
-    from urllib.request import urlopen
-    from urllib.error import URLError
-except ImportError:
     from urllib2 import urlopen
     from urllib2 import URLError
+except ImportError:
+    from urllib.request import urlopen
+    from urllib.error import URLError
 try:
     import cPickle as pickle
 except ImportError:


wjwwood:
Makes sense to me, @tfoote any input?

tfoote:
This is straight from our python2/3 wiki page: http://wiki.ros.org/python_2_and_3_compatible_code#urllib2 

@dirk-thomas is there a reason not to switch the order for this snippet?

efernandez:
For now I've manually reverted the order on a system that regularly calls `rosdep update` and used to fail a lot. So far, it consistently works all the time.

FYI

``` bash
$ python --version
Python 2.7.6
```

``` bash
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 14.04.4 LTS
Release:	14.04
Codename:	trusty
```

``` bash
$ rosversion -d
indigo
```

efernandez:
Looking a little more into this, from http://wiki.ros.org/python_2_and_3_compatible_code (shared by @tfoote) I can see the intention is to try the Python3 package import and fallback to Python2 if it fails:

``` python
try:
    from python3_package import my_name
except ImportError:
    from python2_package import my_name
```

With this PR I'm changing the order of that.

According to https://docs.python.org/2/library/urllib2.html, `urllib2` changed into `urllib.request` and `urllib.error`.

I've just install `ipython3` to check which imports fail, and it seems that this PR is NOT needed, since `urllib.request` fails in Python2 and it falls back to `urllib2` (in Python2 there's `urllib.urlopen`, but not `urllib.request.urlopen`). I'd how to look further into why it fails sometimes. I'd prepare a more exhaustive test. My fault, I should have tested this better. Sorry for the "noise".

PR #446, opened 2016-04-14
strip catkin-sphinx undeclared dependency

tfoote:
It was undeclared and this makes it completely self contained.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -17,7 +17,6 @@ import sys
 sys.path.insert(0, './')
 sys.path.insert(0, '../src')
 
-import catkin_sphinx  # noqa
 from rosdep2 import __version__  # noqa
 
 # If extensions (or modules to document with autodoc) are in another directory,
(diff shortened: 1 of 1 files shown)


wjwwood:
Did you just copy the contents of the `catkin-sphinx` package into here?

tfoote:
@wjwwood this is the theme from catkin-sphinx. Please see https://github.com/ros-infrastructure/catkin-sphinx/issues/6

dirk-thomas:
From my point of view a `-1` on copy-n-pasting this into multiple repositories. If it is being used in more then one place it should stay separately.

An alternative option to consider would be to remove the custom theme altogether and just use the default.

tfoote:
I would be fine with using the default theme too. I just did this to keep
exactly the same behavior. ​

tfoote:
Using default theme

doc/conf.py
```diff
@@ -104,8 +103,8 @@
 
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
-html_theme_path = [os.path.join(os.path.dirname(catkin_sphinx.__file__), 'theme')]
-html_theme = 'ros-theme'
+# html_theme_path = [os.path.join(os.path.dirname(__file__), 'theme')]
+# html_theme = 'ros-theme'
```
dirk-thomas: Please remove obsolete lines instead of commenting them out.

dirk-thomas:
Can you please compare the documentation before and after this patch to ensure that all content is available in the same way.

wjwwood:
-1 to duplicating, -0 to use the default. I think the custom theme is nice to have and I'm willing to spend time preserve it. Please tell what needs to be fixed and I'll make sure it's working.

doc/conf.py
tfoote: I left them commented out to be consistent with the rest of the file. This is a standard template, every other customizable value is commented out throughout the file to show examples of how to do it.

dirk-thomas: I doubt that `ros-theme` is standard though.

PR #449, opened 2016-05-01
Support --quiet option for pip

wkentaro:
On CI testing, the verbose output of pip install is annoying for me.

### DIFF
--- a/src/rosdep2/platforms/pip.py
+++ b/src/rosdep2/platforms/pip.py
@@ -106,12 +106,7 @@ class PipInstaller(PackageManagerInstaller):
         packages = self.get_packages_to_install(resolved, reinstall=reinstall)
         if not packages:
             return []
-        commands = []
-        for pkg in packages:
-            cmd = ['pip', 'install', '-U']
-            if quiet:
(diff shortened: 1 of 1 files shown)


NikolausDemmel:
lgtm

tfoote:
It would be cleaner to create the list and insert or append the quiet argument instead of duplicating the invocation line.

wkentaro:
I have updated the commit.

tfoote:
Thanks that's easier to extend (ala #450) Though you can keep the list comprehension with using the subcommand instead of using a for loop over the packages. Along the lines of:
```
In [1]: subcmd = ['foo', 'bar']
In [2]: pkgs = ['a', 'b', 'c']
In [3]: cmds = [subcmd + [p] for p in pkgs]
In [4]: cmds
Out[4]: [['foo', 'bar', 'a'], ['foo', 'bar', 'b'], ['foo', 'bar', 'c']]
```

wkentaro:
Thanks. That's better.

tfoote:
+1

wjwwood:
Thanks!

PR #458, opened 2016-05-25
Added support for Funtoo linux

allenh1:
This is old. Forgot about it when cleaning up my GitHub account and deleted my fork... So, when I synced it back up I found these extra commits laying around.


allenh1:
Actually, this isn't really necessary. Nevermind. Sorry, everyone.

PR #462, opened 2016-06-13
main: unify output for resolve/db commands, make it parseable

proger:
I also squeezed a tiny cleanup in form of a `resolve_and_print` function.

I'm not sure if the tool output is supposed to be stable, but this PR makes `resolve` and `db` look awk'able (and print the installer by default to save a roundtrip).


src/rosdep2/main.py
```diff
@@ -745,41 +713,57 @@ def command_where_defined(args, options):
...
+                continue
 
-        installer = installer_context.get_installer(rule_installer)
-        resolved = installer.resolve(rule)
-        print("#%s"%(rule_installer))
-        print (" ".join([str(r) for r in resolved]))
+            installer = installer_context.get_installer(rule_installer)
+            resolved = installer.resolve(rule)
+            print("%s:%s:%s"%(rosdep_name, ",".join([str(r) for r in resolved]), rule_installer))
+        except ResolutionError as e:
```
proger: not sure if both `KeyError` and `ResolutionError` handlers are necessary

wjwwood:
Sorry, I know this issue is over a year old.

I don't think we can do this. I do know of at least a few places the output of db is being used. Perhaps we could introduce an optional `--output-version=` option. Where by default it would be something like `plain_text-v1`. And this could be `plain_text-v2` or something like that. We could also have something like` json-vX` in the future which dumped it as some kind of JSON structure. I'm drawing inspiration from `brew info --json=...` on this idea.

Unfortunately, as-is I don't feel comfortable merging this as it departs from the existing out and `rosdep` is not tied to ROS versions so it would affect all versions, past and present.

PR #463, opened 2016-06-21
Remove easy_install from docs

wolfv:
Easy install doesn't appear to exist.


tfoote:
Thanks yes. People should use pip not easy_install.

Issue #274, opened 2013-09-29
rosdep add support for virtual packages


tfoote:
Issues like https://github.com/ros/rosdistro/pull/2100 suggest that it might be useful for rosdep to support OR logic such that it can be resolved by A OR B packages. Or to directly support virtual packages.  

I'm not sure how to use apt exactly to do this.  

Also adding this capability needs to be reviewed carefully to make sure that installations remain deterministically reliable.

ipa-mdl:
This was fixed in #468.  I have tried it with the `linux-kernel-headers` packages introduced in https://github.com/ros/rosdistro/pull/15064

tfoote:
@ipa-mdl thanks for the followup!

tfoote:
I'm reopening this since we have resolved the checking in #468 but from testing on https://github.com/ros/rosdistro/pull/15308 we can't install virtual packages.

ipa-mdl:
My bad, `linux-kernel-headers` seems to work because it has only one candidate.

PR #468, opened 2016-07-19
Fix handle installed virtual packages

kopp:
When a virtual package is requested by a rosdep.yaml file, that was not recognized as being installed because it is not found by `dpkg-query`.  This patch searches for virtual packages using `apt show` and if it's a virtual package searches for the providing actual packages using `apt-cache`.

If `python-apt` (python api package for `apt`) is allowed as new dependency, this could be handled more easily -- see the comment in `_is_installed_as_virtual_package`.

### DIFF
--- a/src/rosdep2/installers.py
+++ b/src/rosdep2/installers.py
@@ -314,6 +314,8 @@ class PackageManagerInstaller(Installer):
     def __init__(self, detect_fn, supports_depends=False):
         """
         :param supports_depends: package manager supports dependency key
+        :param detect_fn: function that for a given list of packages determines
+        the list of installed packages.
         """
         self.detect_fn = detect_fn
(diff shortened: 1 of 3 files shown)


kopp:
This is a first step for #274: When a package providing a virtual package is present, the virtual package is considered installed.  If you try to install a virtual package, apt should make sure that you get some package installed, that provides this virtual package.

tfoote:
This looks valuable. Thanks for the contribution.  

The additional dependency on python-apt might be reasonable. As long as it's available very generally on all platforms in a new enough version.

wjwwood:
Thanks for the contribution @kopp. This looks good as is. My instinct is to merge it as is and track the option to use python-apt as an enhancement.

PR #470, opened 2016-08-30
update yaml-cpp to a valid url. The google code url has changed to 404

tfoote:
We already have this file hosted on download.ros.org


wjwwood:
+1

PR #472, opened 2016-09-06
Arch Linux: Properly invoke pacman with the correct flags.

de-vri-es:
This PR fixes the invocation of pacman.

It addresses the following issues:
* The previous implementation used `pacman -Sy`, which can perform a partial upgrade. This is **very much** not what you should do on Arch Linux [1].
* The `interactive` argument was ignored, meaning pacman would always ask for confirmation and rosdep fails when invoked in non-interactive context.
* Additionally it adds `-q` if `quiet == true`, although I believe this currently does not affect the output of `pacman -S`.

[1] https://wiki.archlinux.org/index.php?title=System_maintenance&redirect=no#Partial_upgrades_are_unsupported

### DIFF
--- a/src/rosdep2/platforms/arch.py
+++ b/src/rosdep2/platforms/arch.py
@@ -38,14 +38,14 @@ PACMAN_INSTALLER = 'pacman'
 
 def register_installers(context):
     context.set_installer(PACMAN_INSTALLER, PacmanInstaller())
-    
+
 def register_platforms(context):
     context.add_os_installer_key(ARCH_OS_NAME, SOURCE_INSTALLER)
     context.add_os_installer_key(ARCH_OS_NAME, PACMAN_INSTALLER)
     context.set_default_os_installer_key(ARCH_OS_NAME, lambda self: PACMAN_INSTALLER)
 
 def pacman_detect_single(p):
-    return not subprocess.call(['pacman', '-Q', p], stdout=subprocess.PIPE, stderr=subprocess.PIPE)    
+    return not subprocess.call(['pacman', '-Q', p], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
 
 def pacman_detect(packages):
     return [p for p in packages if pacman_detect_single(p)]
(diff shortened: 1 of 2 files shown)


de-vri-es:
I forgot:
* `--needed` should not be passed when `reinstall == true`.

I also folded it all into a single pacman invocation, although I'm not sure what the reasoning was for putting it in multiple invocations. A single invocation has the advantage of possibly reducing the number of times hooks are invoked.

wjwwood:
It probably made multiple calls to pacman because that way you can more easily narrow down which rosdep rule was causing which pacman package to get installed. I don't know if that's necessary to preserve or not.

de-vri-es:
It's still the same call to `rosdep` though, I presume? Only `rosdep` would invoke pacman once per package instead of once in total. Regardless, if `pacman` can not find a requested package, it will print a clear error message indicating which package(s) it could not find.

wjwwood:
You don't call rosdep for each rosdep rule you want to resolve, you call it on a package (which has zero to many deps) or many packages. So sometimes it's not clear why rosdep tried to install some particular package.

It's the mapping from the rosdep rule to the pacman command I'm talking about, not whether or not pacman is clear which pacman package is getting installed. Most of the time the rosdep rules mapping are obvious, but not always.

I'm not 100% certain it helps, it's just my guess as to why it used to call it many times. I don't see any problem with making one call instead.

src/rosdep2/platforms/arch.py
```diff
@@ -57,8 +57,14 @@ def __init__(self):
...
         if not packages:
             return []
-        else:
-            return [self.elevate_priv(['pacman', '-Sy', '--needed', p]) for p in packages]
+
+        command  = ['pacman', '-S']
+
+        if not interactive: command.append('--noconfirm')
+        if not reinstall:   command.append('--needed')
+        if quiet:           command.append('-q')
```
wjwwood: nitpick: this repository's files do not follow PEP8, but with additions we can try to follow it where it is not at odds with the files existing style. This isn't PEP8, please put the contents of the if block on new lines.

wjwwood:
Other than a small comment, lgtm.

src/rosdep2/platforms/arch.py
de-vri-es: Fair enough, will do

de-vri-es:
Adjusted for code style. Also removed `#TODO` to implement the `interactive` argument since that is now supported.

wjwwood:
Thanks!

de-vri-es:
Thanks for the quick merge :)

PR #474, opened 2016-09-14
Installers: Never prepend 'sudo' if already root.

### LINKED ISSUE #473: sudo and rosdep
When running ``rosdep`` in a Docker container for CI we always get:

```
Warning: running 'rosdep update' as root is not recommended.
  You should run 'sudo rosdep fix-permissions' and invoke 'rosdep update' again without sudo.
```

Is there a work around?

If not, the debian package ``python-rosdep`` should also require the ``sudo`` package be installed, because currently ``rosdep`` will fail with unuseful error message:
```
rosdep install -r -y -q -n --from-paths . --ignore-src --rosdistro kinetic
ERROR: Rosdep experienced an error: [Errno 2] No such file or directory
rosdep version: 0.11.5
  File "/usr/lib/python2.7/dist-packages/rosdep2/main.py", line 136, in rosdep_main
  File "/usr/lib/python2.7/dist-packages/rosdep2/main.py", line 353, in _rosdep_main
    return _package_args_handler(command, parser, options, args)
  File "/usr/lib/python2.7/dist-packages/rosdep2/main.py", line 447, in _package_args_handler
    return command_handlers[command](lookup, packages, options)
  File "/usr/lib/python2.7/dist-packages/rosdep2/main.py", line 654, in command_install
    installer.install(uninstalled, **install_options)
  File "/usr/lib/python2.7/dist-packages/rosdep2/installers.py", line 485, in install
    verbose=verbose, quiet=quiet)
  File "/usr/lib/python2.7/dist-packages/rosdep2/installers.py", line 535, in install_resolved
    result = subprocess.call(sub_command)
  File "/usr/lib/python2.7/subprocess.py", line 523, in call
    return Popen(*popenargs, **kwargs).wait()
  File "/usr/lib/python2.7/subprocess.py", line 711, in __init__
    errread, errwrite)
  File "/usr/lib/python2.7/subprocess.py", line 1343, in _execute_child
    raise child_exception
OSError: [Errno 2] No such file or directory
```

That is caused by the Docker container not having ``sudo`` installed by default.

NikolausDemmel:
If rosdep is executed as root, we probably never need to prepend a "sudo" to installer commands. This helps in situations where sudo is not installed by default and we typically run as root, eg inside docker images.

Fixes #473.

### DIFF
--- a/src/rosdep2/installers.py
+++ b/src/rosdep2/installers.py
@@ -29,6 +29,7 @@
 
 from __future__ import print_function
 
+import os
 import subprocess
 import traceback
 
@@ -320,7 +321,7 @@ class PackageManagerInstaller(Installer):
         self.detect_fn = detect_fn
         self.supports_depends = supports_depends
         self.as_root = True
-        self.sudo_command = 'sudo -H'
+        self.sudo_command = 'sudo -H' if os.geteuid() != 0 else ""
 
     def elevate_priv(self, cmd):
         """


davetcoleman:
+1 this would be great

PR #476, opened 2016-09-18
archlinux: use pacman -T to verify installation

de-vri-es:
Currently rosdep uses `pacman -Q` to check if a package is installed. However, that does not work with targets provided by a package of a different name.

For example: the `eigen` package also provides the `eigen3` target. Running `pacman -Q eigen3` will exit non-zero though, because there is no package named `eigen3` installed. Running `pacman -T eigen3` will exit with status 0, because the target `eigen3` is provided by an installed package.

This PR changes rosdep to use `pacman -T`.

`pacman -T` prints the non-installed packages, so it could even be used to implement `pacman_detect` with a single pacman invocation. I didn't think that was worth the added complexity over the current implementation though.

### DIFF
--- a/src/rosdep2/platforms/arch.py
+++ b/src/rosdep2/platforms/arch.py
@@ -45,7 +45,7 @@ def register_platforms(context):
     context.set_default_os_installer_key(ARCH_OS_NAME, lambda self: PACMAN_INSTALLER)
 
 def pacman_detect_single(p):
-    return not subprocess.call(['pacman', '-Q', p], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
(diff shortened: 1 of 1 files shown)


hgaiser:
Running `pacman -Q eigen3` does indeed return with exit status 1. Running `pacman -T eigen3` returns with exit status 0 (and just to verify, running `pacman -T eigen33` again returns with exit status 1).

de-vri-es:
I also tested locally. Note that the current behaviour is broken when installing the rosdep key `eigen`, since it resolves to the package `eigen3` on Arch Linux. While in this case it may have been better to have the rosdep key `eigen` resolve to the `eigen` package directly, in general it should be possible to have rosdep keys resolve to virtual targets.

PR #481, opened 2016-10-24
Show installers version with --version option

wkentaro:
Specifically, pip and setuptools version is crucial because
pip distributed on apt is very old on Ubuntu 14.04.
I think showing the version of installer helps user to resolve
their issue caused by old package managing tools.

Usage:

  $ rosdep --version
  source version unknown
  rosdep version 0.11.5
  apt-get version 1.0.1ubuntu2
  pip version 8.1.2
  setuptools version 28.6.1
  gem version 1.8.23

### DIFF
--- a/src/rosdep2/installers.py
+++ b/src/rosdep2/installers.py
@@ -376,6 +376,9 @@ class PackageManagerInstaller(Installer):
         '''
         return not self.get_packages_to_install([resolved_item])
 
+    def get_version(self):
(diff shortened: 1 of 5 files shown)


wjwwood:
I think this is useful, but I cannot accept this as a change to `--version`, since I know people are using `rosdep --version` in scripts and expecting just the version number to come back.

So I opened https://github.com/ros-infrastructure/rosdep/pull/499, but I'll have to decline this one.

PR #482, opened 2016-10-27
install pip first, then install other methods

k-okada:
when we have package that has multiple {run/buidl}_depends with pip and apt (for example), we assume that everything install via apt, then install pip which is not provided from apt
If we install pip first, then pip install everything whcih we assume install via apt

Cc: @wkentaro

### DIFF
--- a/src/rosdep2/installers.py
+++ b/src/rosdep2/installers.py
@@ -487,6 +487,8 @@ class RosdepInstaller(object):
                 previous_installer_key = installer_key
             squashed_uninstalled[-1][1].extend(resolved)
 
+        # sort with priority pip = 100, other 0 (smaller installer runs first)
+        squashed_uninstalled.sort(key = lambda e: {'pip': 100}.get(e[0], 0))
         failures = []
         for installer_key, resolved in squashed_uninstalled:
             try:


tfoote:
Can you explain a little bit more fully what your use case is? Are you saying that you do not have pip installed on the system?

wkentaro:
@k-okada Is this to install apt first, then pip?

for example:

```xml
<run_depend>python-skimage</run_depend>
<run_depend>python-fcn-pip</run_depend>
```

then rosdep install runs below order:

```bash
sudo apt-get install python-skimage
sudo pip install -U fcn
```

wjwwood:
Why does the order matter? If the python dependency needs the apt dependency installed first, then use the `depend` field on the rosdep entry, e.g.:

https://github.com/ros/rosdistro/blob/master/rosdep/osx-homebrew.yaml#L565

wkentaro:
Oh, that is a news for me. Thanks the problem about the order will be resolved with that.

wjwwood:
Sure, I had to add that feature for matplotlib on OS X back in the day 😄.

k-okada:
Let me explain the use case, In the case of

```
<run_depend>python-skimage</run_depend>
<run_depend>python-fcn-pip</run_depend>
```

apt version of `python-skimage` is 0.9.3
```
$ dpkg -l python-skimage
Desired=Unknown/Install/Remove/Purge/Hold
|
Status=Not/Inst/Conf-files/Unpacked/halF-conf/Half-inst/trig-aWait/Trig-pend
|/ Err?=(none)/Reinst-required (Status,Err: uppercase=bad)
||/ Name           Version      Architecture Description
+++-==============-============-============-=================================
ii  python-skimage 0.9.3-4build all          Python modules for image
processi
```

and `python-fcn-pip` is also requires on skimage
```
$  pip show fcn | grep Req
Requires: matplotlib, gdown, scikit-learn, numpy, chainer, plyvel,
scikit-image, tqdm, scipy, h5py
```
and pip version of that is 0.12.3
```
$ pip show scikit-image | grep ^Version
Version: 0.12.3
```

so if we run
```
sudo apt-get install python-skimage
sudo pip install -U fcn
```
then we do not have pip version of skimage, and we'll have 0.9.3 of that,
but if we run
```
sudo pip install -U fcn
sudo apt-get install python-skimage
```
then we'll have 0.12.3 of skimage in our computer.

I think most of ROS tools assuming we have apt installed python code, and
if there is inconsistency in versions, we'll have trouble on this.
For example, I had experienced that rqt did not started because of pip
installed dot_parser (
https://github.com/ros-visualization/rqt_robot_plugins/pull/112#issue-183207356
)
As @wjwwood suggested , using `depend` field is ok, but I think adding
depends to all keys may take so much efforts.


--
◉ Kei Okada


2016-10-28 5:46 GMT+09:00 William Woodall <notifications@github.com>:

PR #483, opened 2016-10-29
Remove checks on keg only homebrew installs

spaghetti-:
Qt5 would otherwise fail on OSX as its a keg only install.

### DIFF
--- a/src/rosdep2/platforms/osx.py
+++ b/src/rosdep2/platforms/osx.py
@@ -184,6 +184,9 @@ def brew_detect(resolved, exec_fn=None):
             pkg_info = json.loads(std_out)
             pkg_info = pkg_info[0]
             linked_version = pkg_info['linked_keg']
+            keg_only = pkg_info['keg_only']
+            if keg_only:
+                return True
             if not linked_version:
                 return False
             for spec in pkg_info['installed']:


wjwwood:
Keg only doesn't really help you though, since it will not be on the default paths... It's the same reason this function returns false if this version is not linked (keg only == not linked).

spaghetti-:
That's true, but how can we get over this issue though? If I'm not wrong for opencv3 theres a ros mirror which gets pulled as a plain cmake package. Qt5 on the other hand is a pain  on OS X particularly because pyside2/shiboken2 (for rqt c++ plugins) does not compile against 5.7.0 (latest on brew). Regardless, they get pulled as keg only.

wjwwood:
Unfortunately, I think you'll have to link the qt5 binaries in manually. Perhaps we could extend `rosdep` to have a "manual action required" message that gets printed until the user does the action, after which it will conclude everything is installed properly.

I don't have time to do that right now, but that's probably our best bet because I don't think it's a great idea to just unlink and link kegs for people. Perhaps that would be ok, but only with additional user confirmations.

wjwwood:
Qt5 is no longer keg-only, and I have doubts about this approach, so I'm going to close this for now. I can reopen it later if necessary.

spaghetti-:
Can verify that there is no issues on installing kinetic with the qt formula out of the box.

PR #491, opened 2016-12-21
Remove build dependency on nose from setup.py

rojkov:
Having nose required to run setup.py is needed only to enable
the `nosetests` command which is redundant since it duplicates
the `test` command. Also it prevents building python-rosdep
in build environments where external downloads during build time
are prohibited in order to provide reproducible and more reliable
builds.

Signed-off-by: Dmitry Rozhkov <dmitry.rozhkov@linux.intel.com>


wjwwood:
lgtm, thanks!

PR #496, opened 2017-01-11
Fix #495

### LINKED ISSUE #495: Can't use rosdep init with local user
Currently it is not possible to setup ros as a non-priviledged user, because '''rosdep init''' will always write to /etc/ros. rospkg provides a function '''get_etc_ros_dir''' that by default uses /etc/ros but will use the contents of the environment variable ROS_ETC_DIR if it is set. However, this call is never executed and its default is used instead.

mightyCelu:
- Use rodpkg.get_etc_ros_dir to get the configuration directory path
- Adapt documentation to include instructions for overriding the
default value

### DIFF
--- a/src/rosdep2/sources_list.py
+++ b/src/rosdep2/sources_list.py
@@ -96,11 +96,7 @@ def get_sources_list_dirs(source_list_dir):
 def get_sources_list_dir():
     # base of where we read config files from
     # TODO: windows
-    if 0:
-        # we can't use etc/ros because environment config does not carry over under sudo
-        etc_ros = rospkg.get_etc_ros_dir()
-    else:
-        etc_ros = '/etc/ros'
+    etc_ros = rospkg.get_etc_ros_dir()
     # compute default system wide sources directory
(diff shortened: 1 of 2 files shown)


src/rosdep2/sources_list.py
```diff
@@ -96,11 +96,7 @@ def get_sources_list_dirs(source_list_dir):
 def get_sources_list_dir():
     # base of where we read config files from
     # TODO: windows
-    if 0:
-        # we can't use etc/ros because environment config does not carry over under sudo
```
wjwwood: It seems this was done on purpose... judging by this comment.

mightyCelu:
Thanks for your feedback.

I created this PR because it was the solution that worked for us when we tried installing ROS as an unpriviledged user. In our setup it was not possible to install it systemwide. Also I think if there is no technical reason that prevents a per-user installation, it should be supported. There will always be users running on computers they can't administer themselves.

There was already an issue pointed out with the commit, I will refine it further.

tfoote:
I'm going to close this as a duplicate of #551 which was merged and then rolled back in #584 The new replacement proposed is in #583

PR #499, opened 2017-02-09
Extension of #481

wjwwood:
I would have pushed directly to #481, but I didn't have permission.

In this extension I:

- restored `--version` to the current behavior
- added `--all-versions` that implements the new behavior
- cleaned up the output formatting
- changed the function name to `get_version_strings` and updated docs to better indicate support for multiple version strings per installer
- added support for Homebrew and Macports

This is what the output looks like on my machine now:

```
william@farl ~
% rosdep --version
0.11.5

william@farl ~
% rosdep --all-versions
0.11.5

Versions of installers:
  Homebrew 1.1.6-45-g9cce341
  Homebrew/homebrew-core (git revision 9c73d; last commit 2017-01-09)
  Macports 2.4.0
  pip 9.0.1
  setuptools 32.3.1
```

### DIFF
--- a/src/rosdep2/installers.py
+++ b/src/rosdep2/installers.py
@@ -376,8 +376,14 @@ class PackageManagerInstaller(Installer):
         '''
         return not self.get_packages_to_install([resolved_item])
 
-    def get_version(self):
-        raise NotImplementedError('subclasses must implement get_version method')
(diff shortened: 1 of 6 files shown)


wkentaro:
+1

PR #505, opened 2017-04-07
Added a new test to show up version checking error

yotabits:
The aim of this test is to show up that rosdep does not care about
the version of the dependencies to be installed in package.xml file.

if rosdep_main function finish with printing:
#All required rosdeps installed successfully
even if package.xml file contains:
depend version_gte="1984.3.14159265359" rospy depend 
the test will be marked as failed.

### DIFF
--- a/src/rosdep2/rospkg_loader.py
+++ b/src/rosdep2/rospkg_loader.py
@@ -117,6 +117,35 @@ class RosPkgLoader(RosdepLoader):
             self._loadable_resource_cache = list(self._rospack.list())
         return self._loadable_resource_cache
 
+
+    def dep_version_checking(self, deps):
+        names_with_version = []
+        for i in (0, len(deps) - 1):
+            version_values = []
+            try:
+                version_values.append(getattr(deps[i], "version_eq"))
+                version_values.append(getattr(deps[i], "version_gt"))
+                version_values.append(getattr(deps[i], "version_gte"))
+                version_values.append(getattr(deps[i], "version_lt"))
+                version_values.append(getattr(deps[i], "version_lte"))
+            except:
+                print("ERROR: Dependencies version checking has failed,")
+                print("if no version tag has been provided this message can be ignored.")
+                return
+
+            for value in version_values:
+                if (value != None):
+                    names_with_version.append(getattr(deps[i], "name"))
+                    break
+
+        if len(names_with_version) > 0:
+           names_with_version = set(names_with_version)
+           print ("WARNING: The following dependencies have a version specified:")
+           for name in names_with_version:
+                print(name)
+                print ("The version tag(s) provided have been ignoredby rosdep")
+
+
     def get_rosdeps(self, resource_name, implicit=True):
         """
         If *resource_name* is a stack, returns an empty list.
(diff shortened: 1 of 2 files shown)


yotabits:
I first tried to make a Pr without "running" rosdep init wich result in a test failure...

Now I added the init  but it's fails because it needs to be run as sudo.
What is the corrections I can make to my test in order to make it pass Travis check? 

Any tip for making this test work is welcome

tfoote:
Can you clarify what your goal is for adding this unit test? rosdep does not support specifying specific versions. This is because it generally relies on the upstream packaged version, especially since most package managers only make one version available.

wjwwood:
To back up what @tfoote said, these versions in the `package.xml` are primarily used during binary packaging to do coordinated upgrades. `rosdep` doesn't use them and that's expected at the moment.

asmodehn:
So far I have noticed two problems (at least with debian packages) : 

- if package A v1.1 upgrade dependency to (incompatible) B `version_gte` 2.0, rosdep will not install package B since the dependency is already satisfied. => A is not working any more.
- if package A v1.1 adds a dependency to C `version_lte`1.9, rosdep will install whatever version is available on the package manager without warning. That can be incompatible C v 2.5 => A is not working any more.

I understand that the version is used when building binary packages, and in that case we can rely on the underlying packaging system to handle that. But this is true only if A is a binary, installed by the packaging system (and we can rely on the distribution to ensure consistency).

Often, A is some source package we are developing/testing on, and we use rosdep to install its dependencies. In that (quite common) use case, rosdep should warn when the version requirements cannot be met by the underlying packaging system, before the user unwillingly install the wrong version of some packages and need to test the whole system, finds out something doesn't work any more, and track down all changes in all packages only to finally realize a wrong version of a package was installed...

The version has been part of the specification for quite some time.
Both REP 127 and REP 140 about `package.xml` (the input for rosdep) mention it :

> All dependencies and relationships may restrict their applicability to particular versions. For each comparison operator an attribute can be used. Two of these attributes can be used together to describe a version range.

Real world deployment can consist of of multiple set of packages (ex: ros distro + company distro + custom development), multiple package repositories (ex: osrf + company stable + company testing) with multiple teams managing them.

Since rosdep is the tool we use to manage ROS packages, it makes sense for that tool to implement the full `package.xml` format specification in order to help users as much as possible. So I think rosdep should parse the version requirement attributes in `package.xml` and check if the version about to be installed matches the requirement. Otherwise there should be an error if a requirement cannot be satisfied, to make the user realize the problem as early as possible.

yotabits:
> So I think rosdep should parse the version requirement attributes in package.xml

Actually the version is parsed, I think a good approach, in first place, would be to let know to the user that the version is ignored if he uses it.

dirk-thomas:
In the case that the version constraint is not satisfied i don't think there is not much `rosdep` can do about it. But I agree with @yotabits and @asmodehn that `rosdep` should print a warning to let the user know about the mismatch.

wjwwood:
Closing in favor of https://github.com/ros-infrastructure/rosdep/pull/506. If this shouldn't be closed please comment here and I can reopen it.

PR #506, opened 2017-04-10
Added version checking error

yotabits:
If a dependency in package.xml use a specific version the user will be
warned that the version is ignored by rosdep.
 
For more precision:
https://github.com/ros-infrastructure/rosdep/pull/505

### DIFF
--- a/src/rosdep2/rospkg_loader.py
+++ b/src/rosdep2/rospkg_loader.py
@@ -117,6 +117,33 @@ class RosPkgLoader(RosdepLoader):
             self._loadable_resource_cache = list(self._rospack.list())
         return self._loadable_resource_cache
 
+    def dep_version_checking(self, deps):
+        names_with_version = []
+        for i in (0, len(deps) - 1):
+            version_values = []
+            try:
+                version_values.append(getattr(deps[i], "version_eq"))
+                version_values.append(getattr(deps[i], "version_gt"))
+                version_values.append(getattr(deps[i], "version_gte"))
+                version_values.append(getattr(deps[i], "version_lt"))
+                version_values.append(getattr(deps[i], "version_lte"))
+            except:
+                print("ERROR: Dependencies version checking has failed,")
+                print("if no version tag has been provided this message can be ignored.")
+                return
+
+            for value in version_values:
+                if (value != None):
+                    names_with_version.append(getattr(deps[i], "name"))
+                    break
+
+        if len(names_with_version) > 0:
+           names_with_version = set(names_with_version)
+           print ("WARNING: The following dependencies have a version specified:")
+           for name in names_with_version:
+                print(name)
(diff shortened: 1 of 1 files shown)


src/rosdep2/rospkg_loader.py
```diff
@@ -117,6 +117,33 @@ def get_loadable_resources(self):
...
+        names_with_version = []
+        for i in (0, len(deps) - 1):
+            version_values = []
+            try:
+                version_values.append(getattr(deps[i], "version_eq"))
+                version_values.append(getattr(deps[i], "version_gt"))
+                version_values.append(getattr(deps[i], "version_gte"))
+                version_values.append(getattr(deps[i], "version_lt"))
+                version_values.append(getattr(deps[i], "version_lte"))
+            except:
```
tfoote: You should catch the specific error that is expected instead of catching all errors.

src/rosdep2/rospkg_loader.py
```diff
@@ -117,6 +117,33 @@ def get_loadable_resources(self):
...
+                print("ERROR: Dependencies version checking has failed,")
+                print("if no version tag has been provided this message can be ignored.")
```
tfoote: If you catch the specific error you won't need this warning about potentially ignoring the message. If there's a known reason for ignoring it we should suppress it in those cases.

yotabits: Actually there is no specific reason for catching an error here, it was just in case the attribute names of object were changed, I think the best option is to remove this "try except", because if attribute name is changed this code should be maintain also. Am I right ?

yotabits:
> I think it would be worth extending this check to try to validate the version available before warning.

I totally agree with you.

> I know this is a bigger change but I don't want to train users to ignore warnings.

I will spend more time digging into the code to do so, as you said user should not get used to ignore warnings
so I will try to slowly code that through different commits with minor changes on that PR.

yotabits:
@tfoote Can I have some feed back on the modifications I already made ? :)

yotabits:
@tfoote  Can I have some feed back on the modifications I already made ? :)

tfoote:
@yotabits I'm travelling at the moment. if you can give a high lever overview of your planned I hopefully can give feedback on that but it will be a little time before I can dig into the code.

tfoote:
@yotabits Reading things through it generally looks reasonable. I would expect that you can find some libraries to to more of the heavy lifting inside the version comparision. My biggest concern is still about Error fatigue. It should not error unless there's something a version missmatch that breaks the constraint. If the constraint is met it's fine to ignore it. And definitely should not produce an error. As there's nothing wrong with the system.

wjwwood:
For everyone's info (and my book keeping), this is waiting on changes requested by @tfoote about the scenarios in which it produces an error and the fact that CI is not passing (CI appears to be passing on master).

Issue #522, opened 2017-07-12
rosdep init error InvalidData: line:


rlk7624:
rosdep version: 0.11.5

Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/rosdep2/main.py", line 136, in rosdep_main
    exit_code = _rosdep_main(args)
  File "/usr/lib/python2.7/dist-packages/rosdep2/main.py", line 351, in _rosdep_main
    return _no_args_handler(command, parser, options, args)
  File "/usr/lib/python2.7/dist-packages/rosdep2/main.py", line 359, in _no_args_handler
    return command_handlers[command](options)
  File "/usr/lib/python2.7/dist-packages/rosdep2/main.py", line 486, in command_init
    data = download_default_sources_list()
  File "/usr/lib/python2.7/dist-packages/rosdep2/sources_list.py", line 325, in download_default_sources_list
    parse_sources_data(data)
  File "/usr/lib/python2.7/dist-packages/rosdep2/sources_list.py", line 366, in parse_sources_data
    raise InvalidData("line:\n\t%s\n%s"%(line, e), origin=origin)
InvalidData: line:
	<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01//EN" "http://www.w3.org/TR/html4/strict.dtd">
type must be one of [yaml,gbpdistro]

tfoote:
This looks like you have invalid data in one of your rosdep sources. It looks like the contents might be empty.

rlk7624:
how to solve this issue?

tfoote:
Make sure that your rosdep sources list is valid. http://docs.ros.org/independent/api/rosdep/html/sources_list.html

You should have something like this: 
```
$ cat /etc/ros/rosdep/sources.list.d/20-default.list 
# os-specific listings first
yaml https://raw.githubusercontent.com/ros/rosdistro/master/rosdep/osx-homebrew.yaml osx

# generic
yaml https://raw.githubusercontent.com/ros/rosdistro/master/rosdep/base.yaml
yaml https://raw.githubusercontent.com/ros/rosdistro/master/rosdep/python.yaml
yaml https://raw.githubusercontent.com/ros/rosdistro/master/rosdep/ruby.yaml
gbpdistro https://raw.githubusercontent.com/ros/rosdistro/master/releases/fuerte.yaml fuerte

[...]

PR #523, opened 2017-07-12
catch download failures better in init, re: #522

tfoote:
I think this will provide a better experience and not tell people that this problem is an internal rosdep issue.

### DIFF
--- a/src/rosdep2/sources_list.py
+++ b/src/rosdep2/sources_list.py
@@ -309,7 +309,8 @@ def download_default_sources_list(url=DEFAULT_SOURCES_LIST_URL):
 
     :param url: override URL of default sources list file
     :return: raw sources list data, ``str``
-    :raises: :exc:`InvalidData`
+    :raises: :exc:`DownloadFailure` If data cannot be
+            retrieved (e.g. 404, bad YAML format, server down).
     :raises: :exc:`urllib2.URLError` If data cannot be
         retrieved (e.g. 404, server down).
     """
@@ -320,9 +321,12 @@ def download_default_sources_list(url=DEFAULT_SOURCES_LIST_URL):
     data = f.read().decode()
     f.close()
     if not data:
-        raise RuntimeError("cannot download defaults file: empty contents")
+        raise DownloadFailure("cannot download defaults file from %s : empty contents" % url)
     # parse just for validation
-    parse_sources_data(data)
+    try:
+        parse_sources_data(data)
+    except InvalidData as e:
+        raise DownloadFailure("Failed to download valid rosdep sources from %s Contents were:{{{%s}}} Parsing error: %s" % (url, data, e))
     return data
 
 def parse_sources_data(data, origin='<string>', model=None):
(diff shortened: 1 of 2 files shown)


src/rosdep2/sources_list.py
```diff
@@ -320,9 +321,12 @@ def download_default_sources_list(url=DEFAULT_SOURCES_LIST_URL):
     data = f.read().decode()
     f.close()
     if not data:
-        raise RuntimeError("cannot download defaults file: empty contents")
+        raise DownloadFailure("cannot download defaults file: empty contents")
```
dirk-thomas: It might be more helpful if the exception raised would contain the `url` for which the problem occured.

src/rosdep2/sources_list.py
```diff
@@ -320,9 +321,12 @@ def download_default_sources_list(url=DEFAULT_SOURCES_LIST_URL):
...
     # parse just for validation
-    parse_sources_data(data)
+    try:
+        parse_sources_data(data)
+    except InvalidData as e:
+        raise DownloadFailure("Failed to download valid rosdep sources from %s Contents were:{{{%s}}} Parsing error: %s" % (url, data, e))
```
dirk-thomas: Since this error has nothing to do with the download but when parsing the data I am not sure this exception type is appropriate.

src/rosdep2/sources_list.py
tfoote: URL added

src/rosdep2/sources_list.py
tfoote: Of the available exceptions this looked the most appropriate to me. If there's one you deem more appropriate I'm happy to switch. 

Although the error may be unrelated to the actual transport, if the downloaded data is invalid or corrupted the place to look is at the network or source. This is following the same pattern as the exceptions in the function `download_rosdep_data` where data downloaded failing validation is considered a failed download. And the DownloadFailed exception is the one that makes sense to add urls to me.

wjwwood:
@tfoote do you have time to look at the conflicts on this branch?

tfoote:
rebased and resolved merge conflict

src/rosdep2/sources_list.py
wjwwood: I kind of agree here, since we could conceivably accidentally make malformed sources and then this error would be incorrectly pointing people to problems with their network rather than pointing them at us.

That being said, I think this is unlikely and most of the time it will be due to network errors. So I'm +0 on adding another error for this case.

@dirk-thomas / @tfoote can you guys pick a new error type or make another argument for leaving it as-is?

NikolausDemmel: > I kind of agree here, since we could conceivably accidentally make malformed sources and then this error would be incorrectly pointing people to problems with their network rather than pointing them at us.

If we are worried about malformed upstream sources, we should add CI in the upstream repo. If it is broken for everyone, people will find the right place to report it even when the error message says its a download failure. I'm +1 on @tfoote's change as is. Or maybe we could rephrase slightly like: `"Failed to download and validate rosdep sources from..."`

tfoote: I've updated the message to be more explicit about the potential causes.

Issue #526, opened 2017-07-19
Support for replace and conflict tag


Baycken:
Hello,
As per [REP 127](http://www.ros.org/reps/rep-0127.html#id69), package.xml now has `replace` and `conflict` tag. I am wondering when those tags will be supported by rosdep.

NikolausDemmel:
What use case do you have in mind? What would you want rosdep to do?

Baycken:
Basically rosdep would skip a key if the key is replaced by another package. If a key is conflicting with a another package, rosdep would also skip the key and warn the user about it.

Baycken:
I have an idea of how to do this and can help code it if needed.

NikolausDemmel:
I don't think there are plans yet to support this, but I'm sure pull requests are welcome if you want to take a stab at implementing this feature.

> Basically rosdep would skip a key if the key is replaced by another package. If a key is conflicting with a another package, rosdep would also skip the key and warn the user about it.

Not that there are many cases to consider here. Just some off the bat:

- what if a rosdep key marked as conflicting is already installed?
- what if a rosdep key makred as conflicting is satisfied by a from-source ros package in your workspace?
- what if there are multiple candidates to be installed that are mutually conflicting? Or if package A requires system dependecy with rosdep A-dep and package B requires ROS package B-dep, which is therefore to be installed (via binary distribution), but now B-dep says it conflicts with A-dep?
- similar situations with `replace` might occur.

I don't want to discourage, just saying that the semantic in various situations should be carefully considered.

Baycken:
@NikolausDemmel  can you have a look at my PR?

NikolausDemmel:
> This could also serve as a more permanent alternative to --skip-keys by allowing packages to blacklist keys of other packages.

Btw, just to add to this, there already is a permanent way to always skip certain keys. See the documentation of `--skip-keys` (http://docs.ros.org/independent/api/rosdep/html/commands.html).

Baycken:
Yeah, I knew that, but that method is machine specific fix and not portable.

wjwwood:
> Yeah, I knew that, but that method is machine specific fix and not portable.

How so?

Baycken:
So you will need to change the base.yaml file to change the definition of that specific package right? So this fix only affect the machine which the `sources.list.d` is modified. While this process can be automated, it's effect still affect that machine only. For organization with many developers or robot deployment, it might be not so good of a solution.

tfoote:
[...]

PR #527, opened 2017-07-19
Implement conflict and replace tag

Baycken:
This is to address #526 

Basic behavior:
- If a package has the replace tag for a key, that rosdep key is skipped.
- If two packages both replace a key, a warning message is printed and the key is skipped

- If a package has the conflict tag for a key. that key is skipped and a warning is generated
- if a package has the conflict tag for a key and that key is already installed, a warning is generated 

It will be up to the user so resolve conflicting replace or conflicts.

### DIFF
--- a/src/rosdep2/catkin_packages.py
+++ b/src/rosdep2/catkin_packages.py
@@ -11,6 +11,8 @@ except ImportError:
     sys.exit(1)
 
 _catkin_workspace_packages = []
+_catkin_replaced_packages = {}
+_catkin_conflicted_packages = {}
 _catkin_packages_cache = {}
 
 
@@ -32,7 +34,21 @@ def find_catkin_packages_in(path, verbose=False):
         return _catkin_packages_cache[path]
     packages = find_packages(path)
     if type(packages) == dict and packages != {}:
-        package_names = [package.name for package in packages.values()]
+        package_names=[]
+        for package in packages.values():
+            package_names.append(package.name)
+            if package.replaces:
+                for replaced in package.replaces:
+                    if replaced.name in _catkin_replaced_packages:
+                        print("WARNING: '{0}' replaces '{1}', but '{1}' is already replaced by '{2}'".format(
+                            package.name, replaced.name,
+                            _catkin_replaced_packages[replaced.name]))
+                    else:
+                        _catkin_replaced_packages[replaced.name] = package.name
+            if package.conflicts:
+                for conflicted in package.conflicts:
+                    if conflicted.name not in _catkin_conflicted_packages:
+                        _catkin_conflicted_packages[conflicted.name] = package.name
         if verbose:
             print("found " + str(len(packages)) + " packages.")
             for package in package_names:
@@ -53,3 +69,13 @@ def set_workspace_packages(packages):
 def get_workspace_packages():
     global _catkin_workspace_packages
     return _catkin_workspace_packages
+
+
+def get_replaced_packages():
+    global _catkin_replaced_packages
+    return _catkin_replaced_packages
+
+
+def get_conflicted_packages():
+    global _catkin_conflicted_packages
+    return _catkin_conflicted_packages
--- a/src/rosdep2/catkin_packages.py
+++ b/src/rosdep2/catkin_packages.py
@@ -40,7 +40,7 @@ def find_catkin_packages_in(path, verbose=False):
             if package.replaces:
                 for replaced in package.replaces:
                     if replaced.name in _catkin_replaced_packages:
-                        print("WARNING: '{0}' replaces '{1}', but '{1}' is already replaced by '{2}'".format(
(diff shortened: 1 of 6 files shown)


src/rosdep2/main.py
```diff
@@ -461,6 +469,20 @@ def _package_args_handler(command, parser, options, args):
                 print("Skipping non-existent path " + path)
         set_workspace_packages(ws_pkgs)
 
+    # Print installed conflicting or replaced packages
+    if command in ['install', 'check'] and (options.ignore_src or options.from_paths):
```
NikolausDemmel: why only if `ignore_src` or `from_paths` is set?

src/rosdep2/main.py
```diff
@@ -461,6 +469,20 @@ def _package_args_handler(command, parser, options, args):
...
+        replaced_pkgs = get_replaced_packages()
+
+        for package in workspace_pkgs:
+            if package in conflicted_pkgs:
+                print("WARNING: conflicting installed packages: '{0}' is conflicting '{1}'!".format(
+                    conflicted_pkgs[package], package))
+            if package in replaced_pkgs:
+                print("WARNING: '{0}' replaces '{1}', but '{1}' is already installed!".format(
+                    replaced_pkgs[package], package))
+
```
NikolausDemmel: You are only checking if the rosdep keys are in the workspace. What about rosdep keys marked as conflicting that are not in your workspace? You are not checking if those are installed.

NikolausDemmel: Also, some packages or system dependencies that are marked as "conflicting" or "replaced" might get installed by the package manager (e.g. as a dependency of another non-conflicting package), so you would really have to check the actual "installed" status for all rosdep keys marked as conflicting or replaced after rosdep is finished.

src/rosdep2/catkin_packages.py
```diff
@@ -32,7 +34,21 @@ def find_catkin_packages_in(path, verbose=False):
...
+                    if replaced.name in _catkin_replaced_packages:
+                        print("WARNING: '{0}' replaces '{1}', but '{1}' is already replaced by '{2}'".format(
+                            package.name, replaced.name,
+                            _catkin_replaced_packages[replaced.name]))
+                    else:
+                        _catkin_replaced_packages[replaced.name] = package.name
+            if package.conflicts:
+                for conflicted in package.conflicts:
+                    if conflicted.name not in _catkin_conflicted_packages:
+                        _catkin_conflicted_packages[conflicted.name] = package.name
```
NikolausDemmel: What about mutually conflicting packages?

NikolausDemmel:
Btw, my comments shouldn't neccessarily mean that they have to be addressed before the PR is merged. They are more general remarks.

What would help is if you could provide a concrete use case that you want to cover with this. Are there any packages that actually use "replaces" and "conflicts" where this is relevant for rosdep? The ideal would be including a unit test that checks this use case (but it is not strictly required I guess).

Baycken:
The main use case will be if a ros package wraps a third party library, for example `libusb` then this package can use the replace tag to tell rosdep to not install `libusb`

src/rosdep2/catkin_packages.py
Baycken: If they are both installed, then a pair of warning will be generated (A conflicts B, and B conflicts A). If only one of them is installed, the other one will be skipped.

Baycken:
Running `rosdep check -a` or `rosdep install -a` would warn user about duplicate replace packages, and/or installed conflicted package. I still need to look into how to make individual package install and check work with replace and conflict.

src/rosdep2/catkin_packages.py
NikolausDemmel: I was thinking of what happens if neither are installed, then neither would get installed. But I guess that cannot even happen, since we are only checking the packages that are already there, and system dependencies (which might not be there yet) cannot specify replaces or conflicts.

NikolausDemmel:
> The incentive for me to work on this is that, arduino-core is a required key for rosserial_arduino. The arduino-core provide by apt-get is really out dated and the arduino-core library is installed to my system via other means . So I would like to have the ability to tell rosdep that a required key is satisfied without needing to add --skip-keys every time, or modify the base.yaml.

Ok, makes sense, thanks.

wjwwood:
> The incentive for me to work on this is that, arduino-core is a required key for rosserial_arduino. The arduino-core provide by apt-get is really out dated and the arduino-core library is installed to my system via other means . So I would like to have the ability to tell rosdep that a required key is satisfied without needing to add --skip-keys every time, or modify the base.yaml.

Could you maybe put in a narrative command line example of this working, something like:

```
$ rosdep check ...
wants to install arduino-core from apt-get
$ "reinstall rosdep with patch"
$ rosdep check ...
new behavior, generates a warning maybe?
```

Just somehow show the before and after workflows that this change enables. That would help me understand what the target is and whether or not this pr achieves it.

Issue #494, opened 2017-01-10
Incorrect dependency install order using manifests


mrath:
We recently switched some of our ROS development from a bunch of custom dependency install (bash) scripts to rosdep. Some of them we created proper .rdmanifest files for, including the `depends` tag. 

However, the installation order doesn't work as expected. At the bottom is an example output, from adding an echo statement to each check presence script. We can see a couple of things from this:

 1. The check presence scripts are executed twice, for some reason
 2. The first time, the order is correct (e.g. `rcw_statistics_database` depends on `lamp`) - the second time it's wrong
 3. When installation begins, it uses the wrong order (the second one)

We're using Ubuntu 14.04.5 and Indigo. Let me know if you need more info.

    $ rosdep install -y -i --from-paths /home/builder/RCW_workspaces/rcw_ci_ws/src
    check-presence-script: ros_patches
    ​check-presence-script: lamp
    check-presence-script: mysql-connector
    check-presence-script: rcw_database
    check-presence-script: robwork
    check-presence-script: rcw_statistics_database
    check-presence-script: smile

    (here it is installing one dependency through apt, that has nothing to do with this)

    check-presence-script: mysql-connector
    check-presence-script: smile
    check-presence-script: rcw_statistics_database
    check-presence-script: lamp
    check-presence-script: rcw_database
    check-presence-script: robwork

    executing command [rosdep-source install http://some_url/mysql_connector.rdmanifest]
    executing command [rosdep-source install http://some_url/smile.rdmanifest]
    executing command [rosdep-source install http://some_url/rcw_statistics_database.rdmanifest]​
    
    (installation fails due to missing dependencies)

NikolausDemmel:
How do you specify the dependencies between the rosdep keys that define the required order?

wjwwood:
Yeah, can you post your rosdep rules file for these custom keys?

mrath:
Can't really post them in full here, since everything is behind a firewall, and the install scripts contains some passwords etc. (don't worry, I know I'm not supposed to do that!).
All the dependencies are set with the `depends` flag in the manifest:
- lamp: no dependencies
- rcw_statistics_database: `depends: [lamp ]`
- rcw_database: `depends: [lamp ]`
- mysql-connector: no dependencies
- smile: no dependencies
- robwork: no dependencies
[...]

PR #529, opened 2017-07-22
Fix depends

wjwwood:
Basically on the second hit of the cache the `dependencies` list would be empty, when it was not to begin with. The list is being consumed (`.pop` until empty) to figure out the next step, so coping it just prevents the cached version from being empty on future cache hits for the same key.

Should fix https://github.com/ros-infrastructure/rosdep/issues/494.


jlack1987:
Did some limited testing with this branch and never ran into the bug reported in https://github.com/ros-infrastructure/rosdep/issues/494

wjwwood:
Thanks for trying it out @jlack1987.

PR #530, opened 2017-07-25
[WIP] Migrate to python-apt

ipa-mdl:
As discussed in #521, this in an implementation with `python-apt`.
Non-debian systems should not be affected at all.
However, the code currently does not handle a missing `apt` module on Debian-based systems properly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ python:
 # command to install dependencies
 install:
 # develop seems to be required by travis since 02/2013
-  - pip install PyYAML argparse rospkg vcstools catkin_pkg python-dateutil rosdistro
+  - pip install PyYAML argparse rospkg vcstools catkin_pkg python-dateutil rosdistro apt
   - python setup.py build develop
   - pip install nose coverage
 # command to run tests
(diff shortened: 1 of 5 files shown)


ipa-mdl:
Looks like `python-apt` cannot be installed with `pip` (and limited effort), the `pypi` version is outdated anyway.

ipa-mdl:
I haven't found any way to install `python-apt` via pip for the non-system python versions from travis.
It works with the ubuntu python version, but requires some system packages to be installed manually.

I could refactor the travis tests to run the tests on system python versions, probably with docker.
And the nosestest are not yet fixed.

@tfoote, @wjwwood: What do you think?

tfoote:
So, after digging in, it appears that python-apt is mostly c++ bindings: https://git.launchpad.net/python-apt/tree/python And it uses the underlying apt implementations which explains why it's not available on non-apt versions. We could do some conditional imports but I think it might make sense to stick with the current system instead of adding the dependency. It's a little bit more code but I think having such a unique snowflake dependency will likely cause more maintenance going forward than the extra logic.

wjwwood:
Ok, so unless @ipa-mdl you disagree, I think we should abandon this approach in favor of #521?

PR #531, opened 2017-07-28
resolve dependencies of meta-packages

### LINKED ISSUE #430: rosdep install fails for metapackages
rosdep install fails to install dependencies from metapackages. 

We tracked down the error to `rospkg_loader.py`:
`get_rosdeps` (https://github.com/ros-infrastructure/rosdep/blob/master/src/rosdep2/rospkg_loader.py#L120) returns an empty list for metapackages. Comment in https://github.com/ros-infrastructure/rosdep/blob/master/src/rosdep2/rospkg_loader.py#L136 is `# stacks currently do not have rosdeps of their own, implicit or otherwise`.

Is this a left-over from pre-catkin and a bug? or is this done by purpose?

Example is the care-o-bot metapackages in https://github.com/ipa320/care-o-bot. Cloning that into an empty catkin workspace and running `rosdep install --from-path src -i -y` does not install any dependencies.

ipa-mdl:
fixes #430 

Haven't tested it yet.
For the sake of simplicity meta-package's `build_depends` and `test_depends` get returned as well, even if these should be empty anyway.

### DIFF
--- a/test/tree/catkin/another_catkin_package/package.xml
+++ b/test/tree/catkin/another_catkin_package/package.xml
@@ -1,0 +1,14 @@
+
+<package>
+  <name>another_catkin_package</name>
+  <version>0.0.0</version>
+  <description>
+    This is a another package
+  </description>
+  <maintainer email="nobody@example.org">Nobody</maintainer>
+  <license>BSD</license>
+
+  <buildtool_depend>catkin</buildtool_depend>
+
+  <build_depend>simple_catkin_package</build_depend>
+</package>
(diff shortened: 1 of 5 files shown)


ipa-mdl:
> Looks reasonable

It wasn't ;)

> but haven't tested either.

I have now added some unit tests and fixed one bug.

Some observations:

* 42c7d927edf0927d59ee5c1575f1a3727aadfdf1 does not affect the stack tests, but is needed for resolving meta-packages
* the implicit flag is ignored for catkin packages

test/tree/catkin/another_catkin_package/package.xml
```diff
@@ -0,0 +1,14 @@
+
```
wjwwood: nitpick: unnecessary empty line

PR #537, opened 2017-08-03
[fix] Exception with multiple lines only prints the first line.

130s:
[20 lines above already handles multiple lines](https://github.com/ros-infrastructure/rosdep/blob/27d372bc551e346ee72d42c21b102d27371bc0d6/src/rosdep2/main.py#L166).

### DIFF
--- a/src/rosdep2/main.py
+++ b/src/rosdep2/main.py
@@ -188,7 +188,7 @@ Please go to the rosdep page [1] and file a bug report with the stack trace belo
 rosdep version: %s
 
 %s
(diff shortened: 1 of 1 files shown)


wjwwood:
When is this an issue? I think the `__str__` of the exception should handle this correctly.

130s:
> When is this an issue?

Unfortunately it's been months since I filed this PR and I don't remember the context, and can't think of situation to raise an exception that expands to multiple lines. But hopefully the title explains enough.

> I think the `__str__` of the exception should handle this correctly.

Ok, commit updated.

PR #542, opened 2017-08-09
choose to install with pip2 or pip3

### LINKED ISSUE #519: Can't choose whether or not to install pip2 or pip3 version of pip package
Have run into the case where I need a specific version of a pip package, but it looks like the in the rosdep yaml files rosdep is not setup to look explicitly for either `pip2` or `pip3` and instead just looks for `pip` and uses whatever version the `pip` executable is using. It would be awesome if we could explicitly say a key should be installed with `pip2 ` or `pip3`

willcbaker:
Fixes #519 

Can't run tests, all tests fail on my machine (even on master), not sure what is required to run nosetests (documentation definitely needs updating)

but should work with some test rosdepkey:

```
python-coverage:
  ubuntu:
    pip:
      packages: [coverage]

python2-coverage:
  ubuntu:
    pip2:
      packages: [coverage]

python3-coverage:
  ubuntu:
    pip3:
      packages: [coverage]
```

### DIFF
--- a/src/rosdep2/platforms/debian.py
+++ b/src/rosdep2/platforms/debian.py
@@ -1,10 +1,10 @@
 #!/usr/bin/env python
 # Copyright (c) 2009, Willow Garage, Inc.
 # All rights reserved.
-# 
+#
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions are met:
-# 
+#
 #     * Redistributions of source code must retain the above copyright
 #       notice, this list of conditions and the following disclaimer.
 #     * Redistributions in binary form must reproduce the above copyright
@@ -13,7 +13,7 @@
 #     * Neither the name of the Willow Garage, Inc. nor the names of its
 #       contributors may be used to endorse or promote products derived from
 #       this software without specific prior written permission.
-# 
+#
 # THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
 # AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
 # IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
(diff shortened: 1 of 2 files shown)


tfoote:
This looks interesting. Since we have side by side installations of the debian python packages it makes sense to support both variants of python via pip.

Does it make sense to support both `pip2` and `pip3` as well as keep the generic `pip`. Clearly we'd need a backwards compatibility but since we're assuming pip is python2 almost implicitly then maybe we should just use that and have `pip3` as the explicit new installer.

I'd like to think a little bit about the implications with respect to systems that may or may not have python2 or python3. The installers should probably check and assert/fail with a useful error message. (And they should also make sure that the version of pip they are running is the right version)

There are some systems that have only python3 or only python2 installed. Arch is the primary example, where they've switched the default to python3. How should this react there? Just assert they can't install python2 packages?

And we'll need to make sure to apply this to all the target platforms not just ubuntu and debian.

emersonknapp:
@tfoote woo this one's been dormant for a long time. I'm interested in this as well, I was attempting to add a rule for https://pypi.org/project/docker/ for python3 and found this - do you remember if you thought further about what we should do?

> since we're assuming pip is python2 almost implicitly then maybe we should just use that and have pip3 as the explicit new installer

I'm inclined to agree with this choice of convention. We'd just need to update the CONTRIBUTING doc to reflect it.

> There are some systems that have only python3 or only python2 installed. Arch is the primary example, where they've switched the default to python3. How should this react there? Just assert they can't install python2 packages?

Also yes, I think, just give an error message if they don't have the correct Python version available.

sloretz:
Related section in ros-infrastructure/rep#201 a proposal for Python 3 support is ROS 1 : https://github.com/ros-infrastructure/rep/pull/201/files#diff-2ae0c7c945887a83beec61fed2c78e8bR113

tfoote:
Thanks for pinging this back. As we've evolved I think that leveraging the ROS_PYTHON_VERSION from the REP will be a better approach. Although this approach is more specific it will cause more overhead for maintaining the rules. And if we're going to have separate keys anyway the python version is somewhat implicitly encoded.

nuclearsandwich:
#709 has been released which uses the ROS_PYTHON_VERSION as discussed above so I'm going to close this PR.

PR #543, opened 2017-08-16
Add __repr__ for SourceInstall

LWisteria:
Because `__rep__` is missing on `SourceInstall`, `--verbose` option doesn't work well for source installer (to display list of installers).


LWisteria:
Thanks for merging!

PR #544, opened 2017-08-16
Fix misdetection on detect_cycles

LWisteria:
If depending children depend on the same package

```
packageA -> packageB -> packageD
         -> packageC -> packageD
```

Recent `detect_cycles` asserts when checking packageD from packageC.

But packageD is not on cycle. This PR fix this problem.


wjwwood:
Thanks for proposing a fix. The CI is currently failing due to this change (imo) and it's not immediately clear to me why this fixes the issue. Can you add a test for this case so it's clear what you're describing is the problem and this fixes it? There are already tests for the dependency graph checker, so you should just be able to extend it:

https://github.com/ros-infrastructure/rosdep/blob/master/test/test_rosdep_dependency_graph.py

LWisteria:
OK, I will try to add tests!

imcelroy:
Hi @LWisteria @wjwwood, we are currently having this issue as well, and it is very critical for us. I can fix/add a unit test for this, and it would be great if we can get this merged asap.

PR #545, opened 2017-08-16
Keep dependencies order

LWisteria:
Because `set` doesn't keep elements order, order of dependency list might be broken.


tfoote:
This is fine, but I'm not sure that we want to rely on ordering of dependencies. Most of the package managers we use don't support ordering. If there's specific ordering required for the installation that ordering should be called out as a dependency.

LWisteria:
I need accurate order from dependency graph when I use source-installer.

PR #551, opened 2017-09-27
Use ROS_ETC_DIR Variable

allenh1:
There is a note in the code that says the etc/ros can't be used because the environment config doesn't carry over under sudo, but I don't think this is actually relevant, as seen from the rospkg code below:

https://github.com/ros-infrastructure/rospkg/blob/124658aa9e7aa5b5db61288624b0aa7858f0db91/src/rospkg/environment.py#L213-L227

---

connects to #495

### DIFF
--- a/src/rosdep2/sources_list.py
+++ b/src/rosdep2/sources_list.py
@@ -96,11 +96,7 @@ def get_sources_list_dirs(source_list_dir):
 def get_sources_list_dir():
     # base of where we read config files from
     # TODO: windows
-    if 0:
-        # we can't use etc/ros because environment config does not carry over under sudo
-        etc_ros = rospkg.get_etc_ros_dir()
-    else:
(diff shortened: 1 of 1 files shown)


wjwwood:
That sounds fine, but why do you need to change this? Or put another way, why is this an issue just now and not in the past?

----

P.S. Please link to code rather than copy-pasting it into the comment.

wjwwood:
Never mind, I see the connected issue now, sorry for the noise.

allenh1:
> That sounds fine, but why do you need to change this? Or put another way, why is this an issue just now and not in the past?

This is just now an issue because I'm installing on a school computer into my scratch directory, for which I do not have root access.

> P.S. Please link to code rather than copy-pasting it into the comment.

Noted.

dirk-thomas:
See #582 for a regression due to this.

Issue #555, opened 2017-10-10
Possible redundant package checks


jacobperron:
It appears the change introduced in #545 results in duplicate package checks.

For example, without the change:

```
$ rosdep install -s --from-paths . -i
dpkg-query: no packages found matching ros-indigo-compressed-depth-image-transport
dpkg-query: no packages found matching ros-indigo-rqt-ez-publisher
dpkg-query: no packages found matching ros-indigo-compressed-depth-image-transport
dpkg-query: no packages found matching ros-indigo-rqt-ez-publisher
#[apt] Installation commands:
  sudo -H apt-get install ros-indigo-compressed-depth-image-transport
  sudo -H apt-get install ros-indigo-rqt-ez-publisher
```

And with the change:

```
$ rosdep install -s --from-paths . -i
dpkg-query: no packages found matching ros-indigo-compressed-depth-image-transport
dpkg-query: no packages found matching ros-indigo-rqt-ez-publisher
dpkg-query: no packages found matching ros-indigo-compressed-depth-image-transport
dpkg-query: no packages found matching ros-indigo-rqt-ez-publisher
dpkg-query: no packages found matching ros-indigo-compressed-depth-image-transport
dpkg-query: no packages found matching ros-indigo-rqt-ez-publisher
dpkg-query: no packages found matching ros-indigo-compressed-depth-image-transport
dpkg-query: no packages found matching ros-indigo-rqt-ez-publisher
dpkg-query: no packages found matching ros-indigo-compressed-depth-image-transport
dpkg-query: no packages found matching ros-indigo-rqt-ez-publisher
dpkg-query: no packages found matching ros-indigo-compressed-depth-image-transport
dpkg-query: no packages found matching ros-indigo-rqt-ez-publisher
dpkg-query: no packages found matching ros-indigo-compressed-depth-image-transport
dpkg-query: no packages found matching ros-indigo-rqt-ez-publisher
dpkg-query: no packages found matching ros-indigo-compressed-depth-image-transport
dpkg-query: no packages found matching ros-indigo-rqt-ez-publisher
dpkg-query: no packages found matching ros-indigo-compressed-depth-image-transport
dpkg-query: no packages found matching ros-indigo-rqt-ez-publisher
dpkg-query: no packages found matching ros-indigo-compressed-depth-image-transport
dpkg-query: no packages found matching ros-indigo-rqt-ez-publisher
dpkg-query: no packages found matching ros-indigo-compressed-depth-image-transport
[...]

PR #556, opened 2017-10-10
Remove redundant dependency checks

jacobperron:
Fix for #555.

This should maintain dependency order (desired from #545) while eliminating duplicate dependencies from the returned list.

@LWisteria, does this negatively impact you?

### DIFF
--- a/src/rosdep2/installers.py
+++ b/src/rosdep2/installers.py
@@ -368,7 +368,8 @@ class PackageManagerInstaller(Installer):
         if not resolved:
             return []
         else:
-            return [x for x in resolved if x not in self.detect_fn(resolved)]
+            detected = self.detect_fn(resolved)
+            return [x for x in resolved if x not in detected]
 
     def is_installed(self, resolved_item):
         '''


jacobperron:
My editor removed all the trailing whitespace from the modified file. If this is undesired, I can undo.

wjwwood:
@jacobperron please avoid them in the future. I'll review this one wi the `?w=1` option on the diff url.

src/rosdep2/installers.py
```diff
@@ -368,7 +368,10 @@ def get_packages_to_install(self, resolved, reinstall=False):
         if not resolved:
             return []
         else:
-            return [x for x in resolved if x not in self.detect_fn(resolved)]
+            detected = set(self.detect_fn(resolved))
+            seen = set()
+            seen_add = seen.add
```
wjwwood: Why do you need to do this? It's the same number of characters.

src/rosdep2/installers.py
```diff
@@ -368,7 +368,10 @@ def get_packages_to_install(self, resolved, reinstall=False):
...
+            return [x for x in resolved if (x not in detected) and (x not in seen or seen_add(x))]
```
wjwwood: What is the `seen` set for? Would this not be simpler:

```python
detected = self.detect_fn(resolved)
return [x for x in resolved if x not in detected]
```

src/rosdep2/installers.py
jacobperron: For execution speed. [Reference](https://stackoverflow.com/a/480227).

Perhaps it is overkill.

src/rosdep2/installers.py
wjwwood: It's fine, I just wasn't aware of why. You might want a comment, otherwise someone might just remove it as redundant.

src/rosdep2/installers.py
wjwwood: It's fine, I just wasn't aware of why. You might want a comment, otherwise someone might just remove it as redundant.

src/rosdep2/installers.py
jacobperron: Yes, it would appear I overthought the problem.

jacobperron:
Alright, simplified and removed the trailing space edits :+1:

PR #557, opened 2017-10-10
Call apt-cache show once with list of packages

### LINKED ISSUE #554: debian.py check for virtual packages is slow
In *debian.py* checking for remaining virtual packages one-by-one in a loop at [L189](https://github.com/ros-infrastructure/rosdep/blob/master/src/rosdep2/platforms/debian.py#L189) is quite slow for large lists of packages. This is due to the `apt-cache show` command. In particular, I'm looking to cut down on build times in some of my pipelines.

I've noticed that calling `apt-cache show` on packages sequentially is a lot slower than passing all packages at once. Example execution times for 58 packages from the shell:

```
time cat test_pkgs | xargs -L1 apt-cache show
```
**1m 12.915s**

```
time apt-cache show $(cat test_pkgs)
```
**0m 2.920s**

I'm exploring possibilities for enhancement of the Python script.

As an aside, I noticed that packages are checked twice over. Does this have something to do with separating build and run depends?

jacobperron:
Attempt to resolve #554.

Instead of checking for virtual packages individually, now the entire list of packages to check are passed to `apt-cache show`. In my tests involving 58 uninstalled packages, this change (in combination with #556) has resulted in a speed up of 3 minutes.

| Version | Time |
|------------|--------|
| 0.11.8    | 3m13.276s |
| this commit | 7m13.761s |
| #556 | 3m7.361s |
| this commit + #556 | 0m8.088s |

It is important to note that I have not had the opportunity to explicitly test with actual virtual packages (ie. the for-loop at [L144](https://github.com/ros-infrastructure/rosdep/compare/master...jacobperron:apt-cache-show-once?expand=1#diff-484efb8c7d4492f49f4b5ebdf0bf5a6aR144) never executes since `virtual_packages` has been empty in my test cases).

### DIFF
--- a/src/rosdep2/platforms/debian.py
+++ b/src/rosdep2/platforms/debian.py
@@ -97,10 +97,10 @@ APT_CACHE_REVERSE_PROVIDE_START_RE = re.compile(
 APT_CACHE_PROVIDER_RE = re.compile('^(.*?) (.*)$')
 
 
-def _is_installed_as_virtual_package(package, exec_fn=None):
+def _is_installed_as_virtual_package(packages, exec_fn=None):
     '''
-    Check whether this is a virtual package and a package providing this
-    virtual package is installed.
+    Check whether these are virtual packages and packages providing them
+    are installed.
 
     :param exec_fn: see `dpkg_detect`; make sure that exec_fn supports a
     second, boolean, parameter.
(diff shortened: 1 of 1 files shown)


ipa-mdl:
This look reasonable, but it will conflict with #521 (as-is).
@jacobperron: Can you please benchmark #521 as well?

jacobperron:
@ipa-mdl I ran the same benchmark test for #521 with the result **0m26.544s**.

ipa-mdl:
Thanks, this looks promising!
However, `apt-cache showpkg` can display for one package only.
Together with #556 it should be fast enough :D

PR #559, opened 2017-10-16
Move from command from pacman to yaourt

patrickelectric:
Mostly packages with tag arch: in [python.yaml](https://github.com/patrickelectric/rosdistro/blob/master/rosdep/python.yaml) are from AUR.
It's necessary to use yaourt or other AUR manager to run rosdep install successfully.

Signed-off-by: Patrick José Pereira <patrickelectric@gmail.com>


tfoote:
> It's necessary to use yaourt or other AUR manager to run rosdep install successfully.

This is a big change. And although we don't actively test with Arch I believe the current implementation has been working. Has Arch had a significant change in infrastructure? I didn't find any articles about that. Also is yaourt truly a drop in replacements? This change is simply changing the command invoked but it's still a 'pacman' registered installer.

patrickelectric:
E.g: [urdfdom-headers](https://aur.archlinux.org/packages/urdfdom-headers/)
[ogre-1.9](https://aur.archlinux.org/packages/ogre-1.9/)
[tango-icon-theme](https://aur.archlinux.org/packages/tango-icon-theme/)
and others..
You can check the current list [here](https://github.com/ros/rosdistro/blob/master/rosdep/python.yaml).
The problem is that ROS need to use AUR packages and I would like to make this possible.
I'll search if is it possible to use pacman to install AUR packages, but I doubt that.
Obs: ros2 works fine :)

patrickelectric:
A good approach would be to ask the user to use yaourt, pacaur or [others](https://wiki.archlinux.org/index.php/AUR_helpers)
To read more about [AUR](https://wiki.archlinux.org/index.php/Arch_User_Repository).

patrickelectric:
```
patrick@dell  ~/ros_ws/kinetic  rosdep install --from-paths src --ignore-src --rosdistro kinetic -y
executing command [sudo -H pacman -S --noconfirm --needed ogre-1.9 python2-pygraphviz python2-empy python2-rosdep python2-catkin_pkg gazebo python2-rospkg urdfdom-headers log4cxx pcl urdfdom tango-icon-theme]
[sudo] password for patrick: 
error: target not found: ogre-1.9
error: target not found: python2-pygraphviz
error: target not found: python2-empy
error: target not found: python2-rosdep
error: target not found: python2-catkin_pkg
error: target not found: gazebo
error: target not found: python2-rospkg
error: target not found: urdfdom-headers
error: target not found: log4cxx
error: target not found: pcl
error: target not found: urdfdom
error: target not found: tango-icon-theme
  pacman: command [sudo -H pacman -S --noconfirm --needed ogre-1.9 python2-pygraphviz python2-empy python2-rosdep python2-catkin_pkg gazebo python2-rospkg urdfdom-headers log4cxx pcl urdfdom tango-icon-theme] failed
```

tfoote:
The first line of the [AUR homepage](https://wiki.archlinux.org/index.php/Arch_User_Repository) says "The Arch User Repository (AUR) is a community-driven repository for Arch users. It contains package descriptions (PKGBUILDs) that allow you to compile a package from source with makepkg and then install it via pacman." 

yaourt is listed in the [AUR Helpers page](https://wiki.archlinux.org/index.php/AUR_helpers) which starts with a warning: "Warning: None of these tools are officially supported by Arch Linux. It is recommended to become familiar with the manual build process in order to be prepared to troubleshoot problems on one's own." and then the first sentence is: "AUR helpers are written to automate certain tasks for the Arch User Repository." Thus I ask again why this change is necessary? And it seems that we should not switch to require any "unofficial" helper tool.

patrickelectric:
@tfoote As you can see in my comment.
**ROS don't work with pacman.**
Why ?
**The packages in rosdep do not exist in core repositories**, it's not possible to install via pacman without scripts or modifications in pacman configuration.
### Why ROS should use yaourt ?
- The "official" manner to install gazebo in arch is via yaourt, [check here](http://www.gazebosim.org/tutorials?tut=install_other_linux&cat=install).

- Yaourt is 100% compatible with pacman.

- You will be one dependence to build the entire system without user interference.

patrickelectric:
If I am completely lost here, please, tell me how do I install ROS in arch linux without running yaourt or modifying rosdep.

tfoote:
We've had community members documenting how to install ROS with the current infrastructure for several years: http://wiki.ros.org/indigo/Installation/Arch and with many contributors. As such I am relatively confident that unless there's been a significant non-backwards compatible change in the Arch packaging system it should basically work, and not require switching the core commands invoked.

You are making a lot of assertions about how it must be done that contradict statements that I've referenced such as the AUR homepage that say that it's designed to work with pacman.

I would suggest that if you'd like to try out ROS on Arch you start by trying to follow these instructions: http://wiki.ros.org/indigo/Installation/Arch that have been maintained by Benjamin Chrétien who has packaged many of the resources necessary, such as gazebo above. And again I'll note that he also states that he's using yaourt for convenience, but it's not necessary. "The current most straightforward way to install ROS Indigo is using the AUR packages that start with ros-indigo. Using a helper such as yaourt or packer, it is possible to install all packages related to ROS Indigo and let the Arch Build System take care of all the dependencies involved." If you'd like to make a fundamental change like switching a core tool to use an "unofficial" tool, it needs to be strongly justified with a reason for the change.

@de-vri-es @bchretien @zootboy @hauptmech @moesenle as Arch users can you comment on this?

patrickelectric:
@tfoote Thank you for the answer, I agree that is possible to install ROS via AUR, but I would like to make ROS itself compatible with arch.

Yes, AUR is compatible with pacman, but not in the way that ROS is working with it, would be necessary to download each package and than running some commands before running pacman directly.
[Install a package from AUR on arch linux](http://blog.spanger.org/?p=1503)
[Arch_User_Repository](https://wiki.archlinux.org/index.php/Arch_User_Repository) Installing packages section

@bchretien, I am doing something wrong running rosdep ? 
My steps:
```bash
yaourt -S python-rosinstall_generator python-rosdep aur/python-wstool
rosdep update
mkdir -p ros_ws/kinetic
cd ros_ws/kinetic
rosinstall_generator desktop_full --rosdistro kinetic --deps --wet-only --tar > kinetic-desktop-full-wet.rosinstall
wstool init -j8 src kinetic-desktop-full-wet.rosinstall
rosdep install --from-paths src --ignore-src --rosdistro kinetic -y
```

zootboy:
I would NAK any changes that swap out pacman for any AUR helper. Yaourt is just one of many. Note that Arch's official policy is that AUR helpers are _not_ supported: 

https://wiki.archlinux.org/index.php/AUR_helpers

Plus, if a TU actually decided to take on ROS packages and pull them into community, using an AUR helper would no longer be correct.

hauptmech:
Some random thoughts:

This patch (and my comments) is in the context of improving the source install method of ROS. 

I'm not sure rosdep has ever functioned as intended on Arch. Anyone?

pacman cannot install AUR packages. This is by design and will never change.

One is already leaving the safety of the official core Archlinux system by installing ROS AUR packages and so I personally see no problem with increasing the functionality of rosdep by using a so called AUR helper. A source install cannot work without an AUR helper.

yaourt is command-line compatible with pacman so the change would work on a technical level. It has a long history. Of the AUR helpers it is a good choice. Or perhaps add a command line option to rosdep to use a custom package manager.

IF ROS becomes an official package, changing from yaourt back to pacman is as trivial as this patch. Though because ROS follows Ubuntu and Arch drops old libraries quicker than Ubuntu, AUR packages will probably always be needed.

@patrickelectric were you able to successfully do a source install of kinetic with this change? If no, what else was needed or where did it die?

Everything zootboy says is correct but I think the pragmatic thing to do is use yaourt if it allows a successful source build.

nuclearsandwich:
This may be out of date given that though I've been an Arch user since 2007, I rarely use yaourt or AUR helpers, but my recollection is yaourt will refuse to run as root or via sudo which means that although it exposes the same CLI as `pacman` it can't be dropped in the way rosdep is using it.

Regardless of workarounds that may be available or changes in yaourt to allow it to run as root, I don't think changing the package installation utility is the best way to resolve the dependency issue on Arch.

I've only come to ROS recently and I haven't plumbed the version control history for context so I don't know what the full story is with respect to AUR packages in the rosdep database for Arch but I agree that trying to target a rolling release from base from a stable software suite will always involve treading off the official path.

I think the best long term solution is to add a separate [rosdep installer](http://docs.ros.org/independent/api/rosdep/html/developers_guide.html#declaring-a-new-installer) (1) for aur packages and either use yaourt to install them, or we could even vendor one of the smaller bash or python2 aur helpers that is less capable but suitable for the kind of use we're after.

(1) hopefully I found the correct documentation for a current version of rosdep 🤞

zootboy:
Rosdep should function for any packages in the official repos, or if someone takes it upon themselves to build a 3rd party binary repo for ROS packages. Rosdep will not, and in my opinion _should not_ function for packages in the AUR. Code from the AUR is not vetted in any way, and literally anyone can create and upload anything they want to it without any oversight. AUR helpers are useful if you know what you're doing and you trust the users that have uploaded the packages you're building, but can easily become foot-guns to the unaware user.

Generally speaking, the Arch policy on foot-guns is that you have to construct them yourselves. If a user wants to swap in an AUR helper in their version of rosdep, they're free to, but I cannot support making that the default.

de-vri-es:
I strongly agree with @zootboy. While I understand that an automatic installer is nice, I don't think it is a good enough reason to downloading and executing unchecked, untrusted code (nothing is).

hauptmech:
@zootboy Yes that is the Arch way, but since you give out foot-guns in the form of the ROS AUR packages you must realize that we are working outside the normal, ideal conditions and a little pragmatism is in order?

@de-vri-es ROS on Arch is exactly that, unchecked, untrusted code. Put a pig in a dress and it's still a pig. Might as well be a useful pig?

I like @nuclearsandwich's idea. Add an ArchAUR rosdep installer.

hauptmech:
I took a closer look at this issue. I see two improvements that could be made.
1) Improve PacmanInstaller to do as much as it can. Right now pacman tries to install the whole list of dependencies all at once and so it's an all or nothing prospect. Modifying this to install packages one at a time means that rosdep would be able to help the user find/install at least some of the dependencies; instead of none which is the current situation.
```diff
+++ src/rosdep2/platforms/arch.py
-    return [self.elevate_priv(command + packages)]
+    return [self.elevate_priv(command + [package]) for package in packages]
```

2) Move AUR packages in the yaml files to another installer as suggested by @nuclearsandwich . `rosdep --os=arch:aur` would probably be suitable. 

Improvement (1) allows rosdep to be useful whenever the database is not up to date, as it is right now and given the frequency at which Arch dumps old libraries, will be often in the future.

(2) means that usage on arch is in keeping with the arch way. The default is safe, but power users are unhindered.

zootboy:
@hauptmech I do like those ideas.

Overall, I'm fine with people using AUR helpers as long as they make the conscious decision to do so themselves (and thus in theory understand the risks). If an AUR "installer" is added to rosdep, I would have the specific helper binary commented out / un-filled, forcing the user to enter their preferred helper.

nuclearsandwich:
I forgot to mention there is some prior art with an AUR installer that uses packer or yaourt. https://github.com/ros-infrastructure/rosdep/pull/276/ It will need re-review which won't make this rosdep release cycle.

patrickelectric:
@nuclearsandwich I can continue #276 if you think that is it possible to merge it.

acxz:
I know this is an outdated issue, but I would just like to clarify one important point that zootboy touched on.

> pacman cannot install AUR packages. This is by design and will never change.

While it is true that pacman cannot install AUR packages, it is possible to use ROS on Arch with pacman.
This can happen through the use of an unofficial user repository: https://wiki.archlinux.org/index.php/unofficial_user_repositories

Once a user adds an unofficial user repository to their Arch Linux, they can use pacman to install packages. Two repositories that contain ros packages include `arch4edu` and `oscloud`. With these repositories it is possible to use rosdep on Arch Linux.

acxz:
@nuclearsandwich in that vein would you be willing to accept PRs adding ros packages to the rosdep list if they are available on an unofficial user repositories such as arch4edu? Since packages in an unofficial user repo can be installed via pacman no backend code will have to change on rosdep's side.
As another point arch4edu is currently the recommend way of installing ROS on Arch based on the current installation page: http://wiki.ros.org/melodic/Installation/ArchLinux

ros-discourse:

https://discourse.ros.org/t/get-ros2-working-on-archlinux/12827/10

PR #566, opened 2018-01-03
Re-add support of FreeBSD to Rosdep

NorwegianRockCat:
This set of changes re-adds support of FreeBSD to Rosdep. The basic files were there, but they used FreeBSD's old package infrastructure and not its [new pkg infrastructure](https://wiki.freebsd.org/pkgng).

I tried to make this use the tests and aside from some assumptions in the tests (i.e., that it expects a debian-based system) they seem to pass. I can add more tests to the FreeBSD installer.

These changes should not be taken to mean that full-on FreeBSD support should be added to ROS, but rather like the attempts on Windows and MacOS, that it should be possible to do some things.

While it's possible to run rosdep now, it doesn't mean that ROS can do anything without corresponding rosdistro. This is to get the ball rolling.

I'm not sure if this is the best way to submit changes. If there is a preferred method, let me know and I'll do that instead. I can be taught.

### DIFF
--- a/src/rosdep2/__init__.py
+++ b/src/rosdep2/__init__.py
@@ -64,10 +64,11 @@ def create_default_installer_context(verbose=False):
     from .platforms import pip
     from .platforms import gem
     from .platforms import redhat
+    from .platforms import freebsd
     from .platforms import slackware
     from .platforms import source
 
-    platform_mods = [arch, cygwin, debian, gentoo, opensuse, osx, redhat, slackware]
+    platform_mods = [arch, cygwin, debian, gentoo, opensuse, osx, redhat, slackware, freebsd]
     installer_mods = [source, pip, gem] + platform_mods
 
     context = InstallerContext()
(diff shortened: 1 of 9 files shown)


tfoote:
Thanks we're always happy to have contributions that help make our tools support more platforms. Pull requests on github are our preferred way for contributions.

Seeing the changes you had to make for the source rosdeps [rep 112](http://www.ros.org/reps/rep-0112.html) They are something that we haven't used for years and we should consider officially deprecating. So for FreeBSD I think it might make sense to start off disabling them so there's no need to modifiy them or do a deprecation cycle since there will be no users of the source rosdeps.

NorwegianRockCat:
OK. Cool, glad I started in the right area. Still not an expert in using Github.

As for the source rosdep, it was only the test that I changed and I only did it because the nosetests ran into it. I guess the best thing would be to add a skip in that test as well, like I did with the debian test. If there's a proper way to do that in nose, I'd happily do it that way.

I can remove the source dependency rosdep from FreeBSD.py file. How would you want me to proceed? Should I send a new pull request without that changes to the source rosdep test?

NorwegianRockCat:
I'm closing this request at the moment, and resubmitting it shortly as an improved patch. This patch has gotten messy and it really isn't that complex. Sorry for the mess.

dirk-thomas:
Please make sure to get the latest state from master since #571 updated the code style with lots of small changes.

NorwegianRockCat:
Will do! That was actually part of the reason I wanted to re-do this.

PR #568, opened 2018-01-07
Fix detection of installed rpms and warn if slow method is being used

CBJamo:
The fallback method for detecting installed rpms (rpm_detect_cmd) was returning a list of all packages passed to it instead of only those installed. On line 92, the .split() was breaking up an error, "no package provides <package>", that rpm outputs. The list 'out_lines' then had an element for every package passed into the function, so the check on line 94 always returns true.

The fix is to instead split on newlines. The rpm error is then a single element in the list, and the string compare on line 94 fails. 

An error message was also added to warn the user that the fallback method is being used if debugging is enabled.

This fix works is for Fedora 27, but this function clearly worked in the past. I'm not sure what changed to produce this bug, but I don't have any older versions of Fedora installed at the moment to test against. I doubt that the fix will cause issues for older versions, however.


wjwwood:
It looks like the RPM tests are now failing on the CI. Please have a look at the failures if you can.

PR #571, opened 2018-01-12
check code style with flake8

dirk-thomas:
The only commits with changes beyond code style:

* 7c61abf3f39afba19b51b1657602158186f1c487
* 367eac32cdc346d3a282cd74da91879d64b5112d
* 70d8ff278812d1c85bbb163bdc4858275228ac36


dirk-thomas:
Since `flake8` doesn't support Python 2.6 anymore (as of version [3.0.0](http://flake8.pycqa.org/en/latest/release-notes/3.0.0.html)) 695bf03a71d0149defca74588b08cf92f929bd1f skips the linter test with Python 2.6.

dirk-thomas:
Finally 63aa66d77b8fbf80cc85bca831e21a8798ffc22a fixes a flaky test which picked one of the ROS distros and checks that some core packages are available. For the case of `melodic` that is not true atm.

Ready for review.

src/rosdep2/installers.py
```diff
@@ -338,15 +342,15 @@ def resolve(self, rosdep_args):
         """
         packages = None
         if type(rosdep_args) == dict:
```
tfoote: The others are switched to use `isInstance` why not the `dict` here too?

dirk-thomas: Because it is not being reported by `flake8`.

dirk-thomas:
> Is there a reason to make this now specifically?

I wanted to contribute another PR and the style in this package was just so obnoxious that I "had to" clean it up in order to stay sane while editing the code. :wink:

PR #572, opened 2018-01-12
[not to be merged] checking Travis status

dirk-thomas:


dirk-thomas:
Actually one of the tests was non-deterministic (picking one of the ROS distros) and checking for existence of some core package. This fails when melodic is chosen since it doesn't have any packages atm... :tired_face:

The latest commit on #571 fixes the test.

PR #573, opened 2018-01-12
add test for wilcard os version

dirk-thomas:
While there are many different cases where `rosdep` could be improved this PR targets one specific case which is fairly common.

Many rosdep keys exist with a generic mapping independent of the OS version. This is great since the rule doesn't have to updated for new OS versions as long as the system package name stays the same. But when at some point the system package name does change there are one two choices with the current version of rosdep which are both not optimal:
* Either all OS versions need to be explicitly enumerated and from that point on forward the entry needs to be updated for every new OS version
* or at some point the "old" OS versions before the system package rename are being dropped to collapse the mapping to a generic rule again (in order to not require to update the entry in the future).

This patch solves this specific problem / use case. It adds special semantic to an OS version named `*`. When a specific rule is being looked up and the desired OS version isn't in the dictionary the logic will now check if an OS version `*` is present. If that is the case the data from `*` is being used (instead of failing the lookup).

The approach is strictly less powerful than allowing to define os version ranges (e.g. "xenial-and-newer"). The advantage is that it doesn't require an order on the os versions and solves at least some current use cases *now*.

If this approach is being approved I am happy to either create a PR against [REP 111](http://www.ros.org/reps/rep-0111.html) or create a new REP describing the delta compared to the existing one.

### DIFF
--- a/src/rosdep2/lookup.py
+++ b/src/rosdep2/lookup.py
@@ -127,9 +127,15 @@ class RosdepDefinition(object):
                     # we've already checked for PACKAGE_MANAGER_KEY, so
                     # version key must be present here for data to be valid
                     # dictionary value.
+                    if (
+                        # if the os_version has the value None
+                        (os_version in data and data[os_version] is None) or
+                        # or if the os_version is not defined and there is no wildcard
+                        (os_version not in data and '*' not in data)
+                    ):
+                        raise ResolutionError(rosdep_key, self.data, os_name, os_version, 'No definition of [%s] for OS version [%s]' % (rosdep_key, os_version))
+                    # if os version is not defined (and there is a wildcard) fallback to the wildcard
                     if os_version not in data:
-                        if '*' not in data:
-                            raise ResolutionError(rosdep_key, self.data, os_name, os_version, 'No definition of [%s] for OS version [%s]' % (rosdep_key, os_version))
                         os_version = '*'
                     data = data[os_version]
                     if type(data) == dict:
--- a/src/rosdep2/lookup.py
+++ b/src/rosdep2/lookup.py
@@ -128,7 +128,9 @@ class RosdepDefinition(object):
                     # version key must be present here for data to be valid
                     # dictionary value.
                     if os_version not in data:
-                        raise ResolutionError(rosdep_key, self.data, os_name, os_version, 'No definition of [%s] for OS version [%s]' % (rosdep_key, os_version))
+                        if '*' not in data:
+                            raise ResolutionError(rosdep_key, self.data, os_name, os_version, 'No definition of [%s] for OS version [%s]' % (rosdep_key, os_version))
+                        os_version = '*'
                     data = data[os_version]
                     if type(data) == dict:
                         for installer_key in installer_keys:
--- a/src/rosdep2/lookup.py
+++ b/src/rosdep2/lookup.py
@@ -127,13 +127,12 @@ class RosdepDefinition(object):
                     # we've already checked for PACKAGE_MANAGER_KEY, so
                     # version key must be present here for data to be valid
                     # dictionary value.
-                    if (
-                        # if the os_version has the value None
-                        (os_version in data and data[os_version] is None) or
-                        # or if the os_version is not defined and there is no wildcard
-                        (os_version not in data and '*' not in data)
-                    ):
+                    # if the os_version is not defined and there is no wildcard
+                    if os_version not in data and '*' not in data:
                         raise ResolutionError(rosdep_key, self.data, os_name, os_version, 'No definition of [%s] for OS version [%s]' % (rosdep_key, os_version))
+                    # if the os_version has the value None
+                    if os_version in data and data[os_version] is None:
+                        raise ResolutionError(rosdep_key, self.data, os_name, os_version, 'Null definition of [%s] for OS version [%s]' % (rosdep_key, os_version))
                     # if os version is not defined (and there is a wildcard) fallback to the wildcard
                     if os_version not in data:
                         os_version = '*'
--- a/test/test_rosdep_lookup.py
+++ b/test/test_rosdep_lookup.py
@@ -165,7 +165,17 @@ def test_RosdepDefinition():
         # tripwire
         str(e)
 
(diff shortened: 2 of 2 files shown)


tfoote:
This will allow it to be more compact, but it will cause overmatching for the other undeclared versions. 
```
2: some key
3: some key
4: other key
5: other key 
6: other key
````
Collapsed to 
```
2: some key
3: some key
*: other key
```
This will overmatch for `1` or any other key.

If we're going to spend time on this it would seem to make sense to roll out targeting the specification proposed for xylem: https://xylem.readthedocs.io/en/latest/spec.html#rules-files using `any_version`. As well as supporting the comma separated versions.

Adding the ordered version list to the OS definitions might not be too much and would avoid the overmatching. I guess we haven't looked at how much more that would take.

**Comment by dirk-thomas** [comment]
As I mentioned above: this patch doesn't aim to be used for keys which require specific os versions to *not be defined*. It basically implements the `any_version` just that the key is names `*` (which could be changed if people think that is better).

The patch specifically avoids the need to have a defined order on the os versions since that requires frequent releases of the tool (for each an every new os version across all supported os name).

The comma separated version doesn't provide any new functionality but only reduces the verboseness of the data.

---

Since exactly zero progress has been made `xylem` for several years I am proposing to take one step in the direction (with this PR) - but to do it *now*. Not when someone will maybe find time to do it "all".

If the current patch prohibits future changes that should absolutely be consider now. But if it successfully provides one part of the desired functionality (with no drawbacks) then I think it shouldn't be hold back.

**Comment by tfoote** [comment]
Yeah, the overhead to implement the ordering does make sense to avoid for now since it will be noteably more effort.

My main suggestion was to use 'any_version' instead of the '*' to move in that direction. The comma separated list is separable, but if we're going to revise the REP it might make sense to bundle both changes.

**Comment by dirk-thomas** [comment]
To cover the case where we want to use the wildcard but do want specific os versions to still be considered as "not available" I added another simple condition in 1c71bf31c265a23454e14af2bb095cbbbcbccefa.

**Comment by wjwwood** [comment]
Doesn't need to be `lucid: null` or `lucid: ~`? On my machine:

```
>>> yaml.load("lucid: None")
{'lucid': 'None'}
```

**Comment by tfoote** [comment]
Setting the null value makes sense to explicitly not match. The canonical null is `~`: http://yaml.org/type/null.html

**Comment by NikolausDemmel** [comment]
+1

This change minimally invasive and provides a notable benefit right away.

I think `*` is also ok as an alternative to to `any_version`, which IIRC is just what we came up with at the time, with no deeper reason to prefer it over something else. Maybe `any_version` more directly conveys that it is a version-wildcard, but it should be quite clear for `*` as well from the context. Also potential extensions to ordered versions can be done with either variant (maybe `'*>=xenial': foo2`). IMHO two criteria for the syntax are important: 

1. intuitively understandable from reading the rules even if you are not familiar with the syntax spec
2. likelihood of clash with potential future version name of some operating system is very low. 

Both fulfill these IMHO (One very minor downsides of `*` I can think of os the need to escape it in YAML).

The extension with explicit Null values is great!

**Comment by wjwwood** [comment]
I guess we need to update the REP before deploying this. So I'll wait to merge it.

**Comment by dirk-thomas** [comment]
> from @tfoote
> The canonical null is `~`

When serializing a `None` value in Python I get `null` (not `~`).

File: src/rosdep2/lookup.py [hunk 1]
```diff
@@ -127,8 +127,16 @@ def get_rule_for_platform(self, os_name, os_version, installer_keys, default_ins
                     # we've already checked for PACKAGE_MANAGER_KEY, so
                     # version key must be present here for data to be valid
                     # dictionary value.
-                    if os_version not in data:
+                    if (
```
130s: IMO it'd be even more clarifying for the users if different error message returns per a) explicitly set `null` v.s. b) N/A-not-defined in rosdep rule, as the reaction users may need to take differs. For (a) users may either give up using rosdep for that dependency or seek source, and for (b) users may add the missing key to the rosdep rule if the rule is simply missing in rule list.

NikolausDemmel: 👍

dirk-thomas: Good point. I separated the conditions and raised two different messages in 4b0685d4623a442194f8f7cc97336733791d16ed.

src/rosdep2/lookup.py
```diff
@@ -127,8 +127,15 @@ def get_rule_for_platform(self, os_name, os_version, installer_keys, default_ins
                     # we've already checked for PACKAGE_MANAGER_KEY, so
                     # version key must be present here for data to be valid
                     # dictionary value.
-                    if os_version not in data:
+                    # if the os_version is not defined and there is no wildcard
+                    if os_version not in data and '*' not in data:
                         raise ResolutionError(rosdep_key, self.data, os_name, os_version, 'No definition of [%s] for OS version [%s]' % (rosdep_key, os_version))
+                    # if the os_version has the value None
+                    if os_version in data and data[os_version] is None:
+                        raise ResolutionError(rosdep_key, self.data, os_name, os_version, 'Null definition of [%s] for OS version [%s]' % (rosdep_key, os_version))
```
130s: I'm afraid `Null definition` and `No definition` (from a [few lines above](https://github.com/ros-infrastructure/rosdep/pull/573/files#diff-6c706391ac7472f4d9badd303bc707f6R132)) may sound the same to majority of users. IIUC, what "Null definition" here means is that there's no way that we get the binary of a certain package (without additional community effort etc.). Do we not want to clearly tell that to the users?

dirk-thomas: Please suggest a specific change which in your opinion would clarify the message.

dirk-thomas:
Thank you for the feedback. Merging and releasing a new minor version.

src/rosdep2/lookup.py
NikolausDemmel: Suggestion: `'[%s] defined as "not available" for OS version [%s]'`

dirk-thomas: Since the PR has already been merged and released please create a new PR for the proposed change.

NikolausDemmel: #581

PR #579, opened 2018-01-23
conflict with upstream's python-rosdep2

mikaelarguedas:
https://github.com/ros/rosdistro/issues/16768#issuecomment-359527357 revealed that the upstream package of rosdep has been renamed to rosdep2 (in artful, bionic and buster). So we need to conflict against that package as well.


mikaelarguedas:
Thanks for the quick review, merging

PR #581, opened 2018-02-07
Tweak error message for Null definition

NikolausDemmel:
Feedback from https://github.com/ros-infrastructure/rosdep/pull/573#discussion_r166764776


dirk-thomas:
Thank you for the improvement.

PR #593, opened 2018-03-12
Eliminate Python 2.6 which is EOL and add flake8

cclauss:
* Remove Python 2.6 because it [went EOL](https://devguide.python.org/#branchstatus) five years ago.
* Add [flake8](http://flake8.pycqa.org) to find Python syntax errors and undefined names.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: python
 python:
+  - "2.6"
   - "2.7"
   - "3.4"
 # command to install dependencies
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,9 +11,9 @@ install:
   - pip install nose coverage flake8
 before_script:
   # stop the build if there are Python syntax errors or undefined names
-  - flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,6 +1,5 @@
 language: python
 python:
-  - "2.6"
```
wjwwood: Yeah, so this still here because we still release new binaries for several EOL Ubuntu distributions that use 2.6:

https://github.com/ros-infrastructure/rosdep/blob/master/stdeb.cfg#L6

We could, and probably should, just drop support for them, but since rosdep parses files hosted online, if we were to make a change to the hosted file's syntax then rosdep on the older platforms would just break (since there's no newer version of rosdep for them). The work around would be that those users would need to manually install an older rosdep from pip or source, but then if we stop testing for 2.6 then they might have to also patch rosdep to support their use case.

So that's why it's lasted until now, specifically for rosdep. For any tool that worked off local files, this would be an easy decision.

----

So we have two routes here, if you remove 2.6 here then you should also update the pr to prune the list of suites we support that also need 2.6 support. However, I'll need to send a message to discourse letting people know and give them a period of time to respond.

If not, you can remove this line from the pr and I can merge it now.

wjwwood:
Looks like flake8 isn't working on 2.6, you could just disable the flake tests on that version, since we're running the same checks in the other versions we're checking.

PR #595, opened 2018-03-15
Add badges for build status and code coverage

Nickolaim:
The badge for build status will start working automatically, since it takes status from Travis CI.

For the code coverage the repo owner needs to login to [https://codecov.io/gh](https://codecov.io/gh) and enable code coverage for `ros-infrastructure/rosdep`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ install:
   - python --version
   - pip install PyYAML argparse rospkg vcstools catkin_pkg python-dateutil rosdistro
   - pip install -e .
-  - pip install nose coverage flake8 mock
+  - pip install nose coverage flake8 mock codecov
 before_script:
   # stop the build if there are Python syntax errors or undefined names
   - if [ "${TRAVIS_PYTHON_VERSION}" != "2.6" ] ; then flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics; fi
@@ -32,5 +32,7 @@ before_script:
 # command to run tests
 script:
   - nosetests --with-coverage --cover-package=rosdep2 --with-xunit test
+after_script:
+  - codecov
 notifications:
   email: false
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 rosdep
 ------
-[![Build Status](https://travis-ci.org/ros-infrastructure/rosdep.svg?branch=master)](https://travis-ci.org/ros-infrastructure/rosdep)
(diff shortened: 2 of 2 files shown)


Nickolaim:
The fact that the change that basically updates README.md file got a build break is terrible. And it failed on mac that I worked before.  Interesting that this is the first time I hit a build break like this, I've done several dozens of builds.  I will continue looking for the health of mac builds over time.

As for the change, I'll do a small, hopefully, useful change that will trigger another build.  Hope it is going to fix status from red to green.

Nickolaim:
This is how README.md is going to look after the change.  Code coverage badge is not yet filled because there is no data yet.


screen shot 2018-03-16 at 8 48 44 pm

Nickolaim:
Hey, any feedback on this request?

NikolausDemmel:
The build break it seems was spurious. It failed to download a file from github, which can happen from time to time.

This change seems helpful.

+1

Nickolaim:
Absolutely - if something does not work, it needs to be taken out :)

Nickolaim:
I cannot find the reported behavior - [https://codecov.io/gh/ros-infrastructure/rosdep/commits](https://codecov.io/gh/ros-infrastructure/rosdep/commits) shows that code coverage is on 75.24% or higher.

dirk-thomas:
The quotes text was copy-n-pasted from the status messages on the GitHub PR coming from each CI run.

Nickolaim:
Interesting.  codecov produces coverage.xml for every Travis CI job.  Logs for the code coverage with coverage xml for #609 available at [this page](https://codecov.io/gh/ros-infrastructure/rosdep/commit/1479f2663a1bd168c9d3a21afc05b26d8d0b4ae8/build).

I searched for `line-rate=` and results differ per job:
- 0.7441 - osx 3.6.4
- 0.7427 - linux 3.4, linux 3.5, linux 3.6
- 0.7417 - osx 2.7.14, osx
- 0.7406 - linux 2.7

According to [codecov documentation](https://docs.codecov.io/docs/merging-reports) the tool always merge results from multiple languages, I assume the same works for multiple jobs.

With all above it is not clear to me why 75.03% was reported at some point.  My guess is when only few jobs were completed, and their merge results was 75.03% a premature notification was sent.  The number was adjusted when data from all jobs got uploaded.

For preventing such cases in the future one might consider uploading code coverage results only for one job.

Issue #596, opened 2018-03-17
flake8 executed in 2 different ways with different settings


Nickolaim:
rosdep has flake8 unit test (see [https://github.com/ros-infrastructure/rosdep/blob/master/test/test_flake8.py](https://github.com/ros-infrastructure/rosdep/blob/master/test/test_flake8.py)]

Pull Request  #593 also run flake8 but in Travis CI with a different set of options.

What's the value of having flake8 being run in 2 different ways?  Running them from unit test seems like the best option - unit tests not only executed in Travis CI but on local dev machines.

@cclauss , @wjwwood - you might have an opinion.

I am proposing to use one approach and merge the rules; I personally prefer the unit test.  Happy to do the change.

cclauss:
I have no objection to this course of action.  I agree that running flake8 in two different locations is suboptimal.

wjwwood:
Oh, I didn't notice (we have a flake8 automated test on some repositories but not others.

I have a slight preference for the automated test, rather than running it explicitly in travis, simply because then that test is run locally when running tests too.

Pull requests to unify this would be appreciated.

Nickolaim:
I have a fix for this.  Will do a PR later today.

Sent from a mobile device.
________________________________

PR #599, opened 2018-03-20
Remove flake from .travis.ci, since it is already running in unit tests.

Nickolaim:
Move the only missing `max-complexity` option from .travis.yml to
the unit test. Setting max line length to 127 caused 31 errors -
not practical right now, keep at the old value of 200.
The other rules are already enabled. This is the list of rules enabled
in unit tests: ['E', 'F', 'W', 'C90'].

Fixes issue #596


wjwwood:
Re: codecov I tried triggering a manual build of master: https://travis-ci.org/ros-infrastructure/rosdep/builds/355689951

I'll then restart the CI on this pr to see if coverage is calculated correctly.

PR #601, opened 2018-03-22
Fix for #598 - better handling invalid packages

### LINKED ISSUE #598: improve error message when package.xml is malformed
Anytime there's an issue with a `package.xml` you get something like reported in https://github.com/ros-infrastructure/rosdep/issues/597#issue-306498319.

Instead it should avoid the traceback and the message to report it on the rosdep repository. Also, it should include which `package.xml` has the issue.

Nickolaim:
This change with [https://github.com/ros-infrastructure/catkin_pkg/pull/209](https://github.com/ros-infrastructure/catkin_pkg/pull/209) do the following:
- Add a file name to `InvalidPackage` exception, so it is easier to find out what's wrong
- Do not print the generic error message for this exception, since it is pretty much clear what's wrong

Fixes #598

### DIFF
--- a/src/rosdep2/main.py
+++ b/src/rosdep2/main.py
@@ -73,6 +73,7 @@ from .rosdistrohelper import PreRep137Warning
 from .catkin_packages import find_catkin_packages_in
 from .catkin_packages import set_workspace_packages
 from .catkin_packages import get_workspace_packages
+from catkin_pkg.package import InvalidPackage
 
 
 class UsageError(Exception):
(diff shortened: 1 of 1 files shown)


Nickolaim:
Any suggestions on what else should be done?  Adding test for 2 lines of code does not look reasonable.

Nickolaim:
[https://github.com/ros-infrastructure/catkin_pkg/pull/209](https://github.com/ros-infrastructure/catkin_pkg/pull/209) has been merged.  It addressed:

> Also, it should include which package.xml has the issue.

This PR addresses:

> Instead it should avoid the traceback and the message to report it on the rosdep repository.

Nickolaim:
Did not generate enough interest for the branch owners.  Cleaning up.

NikolausDemmel:
@Nickolaim, sorry, I guess maintainers are just really busy with other stuff.

@wjwwood I believe this is good to merge as is and should help to prevent some pointless issues from being opened in the future. I suggest to reopen and merge.

tfoote:
It does look like an improvement. We do have a lot of other things going on at the moment. I would suggest that adding a very simple unit test for it would be worth it. It's adding another branch point in the logic and should be relatively easy to test just adding one case. The test coverage is not so much about making sure that it works now, but that it continues to work as things evolve in the future.

PR #612, opened 2018-06-06
fix quiet mode of debian installer

ipa-mdl:


tfoote:
Good catch, I updated it for a slightly more compact implementation.

ipa-mdl:
The same structure is used in `redhat.py`, but without the typo.
The test failure is unrelated.

wjwwood:
@tfoote should I merge this then (you can feel free to merge it as well)? I don't have time to do a release right now. Though a rosdep release is something I should do soon, maybe just after the ROS 2 release.

ipa-mdl:
> The failure mode for this is that it won't pass the quiet flags through if you use -y

The quiet flag was not passed at all.
This bug was present for 4 years, so a release is not urgent.

708yamaguchi:
@tfoote @wjwwood 
Could you please release `0.12.3` version of `rosdep`?
I have a problem that our `travis` fails because the travis log exceeds its maximum log length. (for example https://github.com/jsk-ros-pkg/jsk_robot/pull/963 and https://github.com/jsk-ros-pkg/jsk_robot/pull/969)
I think this issue will be solved if this commit is merged.

ipa-mdl:
The last release is more than 7 months old..

nuclearsandwich:
This pull request has been released in rosdep 0.13.0. Thanks for contributing!

PR #613, opened 2018-06-22
Add reference to LICENSE file in stdeb.cfg

tfoote:
Add the parameter to install the license/copyright file in the debian way.
https://github.com/astraw/stdeb#stdebcfg-configuration-file


tfoote:
Well it does drop the file into that location, but doesn't format it as such. Would it make sense to just update the LICENSE file to use that format?

nuclearsandwich:
> Well it does drop the file into that location, but doesn't format it as such. Would it make sense to just update the LICENSE file to use that format?

oh hmm. It doesn't actually appear to be mandatory that the copyright file adhere to the Debian machine readable format (described [here](https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/)). A good amount of sampled copyright files from official packages are unformatted. So this is fine as is.

Issue #379, opened 2015-04-06
Python 3 rosdep database is incompatible with Python 2


cottsay:
I (accidentally) used a Python 3 version of rosdep to run `rosdep update`. I then tried releasing a package in a Python 2 version of Bloom, which called into the rosdep API, and threw this:

```
Traceback (most recent call last):
  File "/usr/bin/git-bloom-generate", line 9, in <module>
    load_entry_point('bloom==0.5.19', 'console_scripts', 'git-bloom-generate')()
  File "/usr/lib/python2.7/site-packages/bloom/commands/git/generate.py", line 252, in main
    run_generator(generator, args)
  File "/usr/lib/python2.7/site-packages/bloom/commands/git/generate.py", line 127, in run_generator
    gen.pre_modify)
  File "/usr/lib/python2.7/site-packages/bloom/commands/git/generate.py", line 101, in try_execute
    retcode = func(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/bloom/generators/rpm/generator.py", line 540, in pre_modify
    while not self._check_all_keys_are_valid(peer_packages):
  File "/usr/lib/python2.7/site-packages/bloom/generators/rpm/generator.py", line 511, in _check_all_keys_are_valid
    retry=False)
  File "/usr/lib/python2.7/site-packages/bloom/generators/common.py", line 137, in resolve_rosdep_key
    view = get_view(os_name, os_version, ros_distro)
  File "/usr/lib/python2.7/site-packages/bloom/generators/common.py", line 79, in get_view
    value = get_catkin_view(ros_distro, os_name, os_version, False)
  File "/usr/lib/python2.7/site-packages/rosdep2/catkin_support.py", line 108, in get_catkin_view
    sources_loader = SourcesListLoader.create_default(matcher=sources_matcher)
  File "/usr/lib/python2.7/site-packages/rosdep2/sources_list.py", line 607, in create_default
    sources = load_cached_sources_list(sources_cache_dir=sources_cache_dir, verbose=verbose)
  File "/usr/lib/python2.7/site-packages/rosdep2/sources_list.py", line 508, in load_cached_sources_list
    return parse_sources_data(cache_data, origin=cache_index, model=model)
  File "/usr/lib/python2.7/site-packages/rosdep2/sources_list.py", line 366, in parse_sources_data
    raise InvalidData("line:\n\t%s\n%s"%(line, e), origin=origin)
InvalidData: line:
[...]

PR #633, opened 2018-09-30
Use pickle protocol format version 2.

nuclearsandwich:
Proposed as a fix for #379.

So far this only #worksonmymachine. Following the queues from @NikolausDemmel in [this comment](https://github.com/ros-infrastructure/rosdep/issues/379#issuecomment-90207509) I read through the docs and tried to use the cache created in python3 from python2 without the fix_imports flag and was successful so I did not apply it.

I have not added any tests of the unpickling. I'm not sure how effective such a test would be with a static blob and I don't know if we want to install python3 separately in all of the python2 environments.


NikolausDemmel:
Cool, thanks for looking into this.

The patch looks reasonable. Also, `fix_imports` seems to be True by default, so we are using it.

Have you also tried reading a dump created with python 2 in python 3? One thing I could see going wrong is that after loading in python 3 strings are now bytes instead, but maybe not.

Some sort of test would be great for future regressions. A test that could run in just the current version that is being tested (be it pyhton 2 or 3):

- write small cache and compare blob to checked in reference blob (I'm assuming the binary representation is not changing, which would have to be verified).
- load from the checked in blob and do simple query

Except for testing aspect, +1 on this PR

cottsay:
Relevant documentation for observers: https://docs.python.org/3/library/pickle.html#pickle.dump

Looks like the correct and least disruptive fix to me. It's nice that the docs specifically call out Python2 compatibility.

NikolausDemmel:
I tested this branch in both directions (updating with py2, reading with py3 and vice versa), and it seems to work fine, where master fails if py2 tries to read a py3 created cache.

+1 for merging as is.

nuclearsandwich:
@NikolausDemmel I have a branch that tests unpickling a file created with this branch in python2 and python3 produces equivalent data structures I just have to find time to get it cleaned up and pushed.

wjwwood:
Looks like the only failure is related to new warnings from flake8, so this should be ok to merge. It would be nice to fix those in a different pr.

NikolausDemmel:
> Looks like the only failure is related to new warnings from flake8, so this should be ok to merge. It would be nice to fix those in a different pr.

There is already #636 ;-)

NikolausDemmel:
No worries, I can relate. Thanks for merging. :-)

PR #638, opened 2018-10-29
Make DNF install best-effort like YUM

cottsay:
When the `DnfInstaller` was added, I believed that DNF's default behavior was best-effort, and this was the reason that the YUM option `--skip-broken` didn't work. It turns out that this is only the case for updates, and that the `install` verb still needs to be told to be best-effort.

At some point `--skip-broken` was added to DNF as an alias for `--setopt=strict=0`. Since older DNF versions don't support `--skip-broken`, we should use `--setopt=strict=0` directly to achieve the expected behavior here.

When a broken package install is attempted, DNF will return 0 and rosdep will detect the missing package as "was supposed to be installed but was not," prompting further inspection of the DNF console output.

### DIFF
--- a/src/rosdep2/platforms/redhat.py
+++ b/src/rosdep2/platforms/redhat.py
@@ -155,13 +155,13 @@ class DnfInstaller(PackageManagerInstaller):
         if not packages:
             return []
         elif not interactive and quiet:
-            return [self.elevate_priv(['dnf', '--assumeyes', '--quiet', 'install']) + packages]
+            return [self.elevate_priv(['dnf', '--assumeyes', '--quiet', '--setopt=strict=0', 'install']) + packages]
         elif quiet:
-            return [self.elevate_priv(['dnf', '--quiet', 'install']) + packages]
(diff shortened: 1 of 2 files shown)


cottsay:
@richmattes FYI

cottsay:
Relevant docs: https://dnf.readthedocs.io/en/latest/cli_vs_yum.html#skip-broken

cottsay:
Looks good now, @nuclearsandwich. Thanks!

nuclearsandwich:
@cottsay I know we discussed this briefly offline. But for my benefit this change updates the dnf installer to match the behavior of the apt installer when installing multiple packages correct?

cottsay:
I believe so, yes.

#214  where this change was made to Yum.

My understanding is that best-effort is desired here, and Apt doesn't (or at least didn't) have that functionality, so it is invoked individually for each dependency. When DNF was added, we *thought* that behavior of YUM became the default, but that was only true for the `update` verb, supposedly. So it definitely acts differently from Apt and YUM right now.

PR #647, opened 2019-01-10
skip EOL distros by default, add option to include them

dirk-thomas:
With the increasing number of already EOLed ROS distros the time for `rosdep update` increases with it.

Currently `groovy`, `hydro` and `jade` are EOLed ROS 1 distros. With the merge of the ROS 2 distro into the same repo `ardent` will be another EOLed one.

This patch changes the default behavior of `rosdep update` to not fetch / consider EOLed distros which will improve its performance. The new option `--include-eol-distros` can be used to continue including EOLed distros.

### DIFF
--- a/src/rosdep2/sources_list.py
+++ b/src/rosdep2/sources_list.py
@@ -439,7 +439,8 @@ def _generate_key_from_urls(urls):
 
 
 def update_sources_list(sources_list_dir=None, sources_cache_dir=None,
-                        success_handler=None, error_handler=None):
+                        success_handler=None, error_handler=None,
+                        include_eol_distros=False):
     """
     Re-downloaded data from remote sources and store in cache.  Also
     update the cache index based on current sources.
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Two of the Travis builds fail for unrelated reasons and are already present on the master branch.

src/rosdep2/sources_list.py
```diff
@@ -439,7 +439,8 @@ def _generate_key_from_urls(urls):
 
 
 def update_sources_list(sources_list_dir=None, sources_cache_dir=None,
-                        success_handler=None, error_handler=None):
+                        success_handler=None, error_handler=None,
+                        include_eol_distros=False):
```
jacobperron: The new parameter can be added to the docstring.

dirk-thomas: Good point: f1ba42dc68f895dadcc18350d228eb62cff01854

dirk-thomas:
Added a note to the wiki:
* http://wiki.ros.org/action/diff/rosdep?action=diff&rev1=87&rev2=88
* http://wiki.ros.org/action/diff/ROS/Tutorials/rosdep?action=diff&rev1=10&rev2=11

dirk-thomas:
@wjwwood FYI. I will go ahead and merge and release this change (even though CI fails).

ipa-mdl:
> Added a note to the wiki:

@dirk-thomas: This change only affects EOL systems (incl. CI ;)), so a post on ROS discourse might be more suitable to reach these legacy users.
Especially w.r.t `indigo`, which will be EOL soon.

dirk-thomas:
> This change only affects EOL systems (incl. CI ;)), so a post on ROS discourse might be more suitable to reach these legacy users.

https://discourse.ros.org/t/rosdep-and-eol-distros/7640

PR #651, opened 2019-01-23
use yaml.safe_load for untrusted yaml input

dirk-thomas:
The patch only updates the invocation where untrusted sources are being processed.


dirk-thomas:
I will leave it up to the maintainer(s) to merge and release with the CI failing.

Issue #645, opened 2018-12-04
`rosdep install` won't work on windows


yyu:
`C:\dev>` `\Python37\python.exe \Python37\Scripts\rosdep install --from-paths src --ignore-src --rosdistro bouncy -y --skip-keys "console_bridge fastcdr fastrtps libopensplice67 rti-connext-dds-5.3.1 urdfdom_headers"`

```
Please go to the rosdep page [1] and file a bug report with the stack trace below.
[1] : http://www.ros.org/wiki/rosdep

rosdep version: 0.13.0

Traceback (most recent call last):
  File "C:\Python37\lib\site-packages\rosdep2\main.py", line 140, in rosdep_main
    exit_code = _rosdep_main(args)
  File "C:\Python37\lib\site-packages\rosdep2\main.py", line 391, in _rosdep_main
    return _package_args_handler(command, parser, options, args)
  File "C:\Python37\lib\site-packages\rosdep2\main.py", line 446, in _package_args_handler
    pkgs = find_catkin_packages_in(path, options.verbose)
  File "C:\Python37\lib\site-packages\rosdep2\catkin_packages.py", line 33, in find_catkin_packages_in
    packages = find_packages(path)
  File "C:\Python37\lib\site-packages\catkin_pkg\packages.py", line 89, in find_packages
    packages = find_packages_allowing_duplicates(basepath, exclude_paths=exclude_paths, exclude_subspaces=exclude_subspaces, warnings=warnings)
  File "C:\Python37\lib\site-packages\catkin_pkg\packages.py", line 139, in find_packages_allowing_duplicates
    pool = multiprocessing.Pool()
  File "C:\Python37\lib\multiprocessing\context.py", line 119, in Pool
    context=self.get_context())
  File "C:\Python37\lib\multiprocessing\pool.py", line 177, in __init__
    self._repopulate_pool()
  File "C:\Python37\lib\multiprocessing\pool.py", line 238, in _repopulate_pool
    self._wrap_exception)
  File "C:\Python37\lib\multiprocessing\pool.py", line 257, in _repopulate_pool_static
    w.start()
  File "C:\Python37\lib\multiprocessing\process.py", line 112, in start
    self._popen = self._Popen(self)
  File "C:\Python37\lib\multiprocessing\context.py", line 322, in _Popen
    return Popen(process_obj)
  File "C:\Python37\lib\multiprocessing\popen_spawn_win32.py", line 33, in __init__
[...]

PR #656, opened 2019-01-28
[Windows] Add console script entry point

seanyen:
Add console script entry points for platforms (e.g. Windows) not supporting shebang to run the Scripts.


src/rosdep2/main.py
```diff
@@ -347,7 +347,7 @@ def _rosdep_main(args):
         installer_keys = get_default_installer()[1]
         version_strings = []
         for key in installer_keys:
-            if key is 'source':
+            if key == 'source':
```
seanyen-msft: Flake8 fix for `F632 use ==/!= to compare str, bytes, and int literals`.

seanyen:
@nuclearsandwich @wjwwood Ping! it would be nice I can get some feedbacks on this change, and let me know anything needed to be changed.

And btw, I think this can address this issue: https://github.com/ros-infrastructure/rosdep/issues/645

nuclearsandwich:
I will have to check, as I don't know off the top of my head, how far back the entry_points API is supported by setuptools. There may be platforms that can't make use of them yet which rosdep is still released for.

seanyen:
Run a quick check, `console_scripts` was there in setuptools even when it is still in a pre-release state (0.6a2, which was dated back to 2005.) Given how long it has been exercised by other projects\platforms, presumably it should be matured enough not to give people suprise. Hope that gives some more confident from other perspectives. 😃  

https://setuptools.readthedocs.io/en/latest/history.html#a2

seanyen:
@nuclearsandwich Just checking back any feedbacks on this change?

seanyen:
@sloretz Was it something we could consider to merge?

seanyen:
@sloretz @dirk-thomas Happy new year. This is ready for review and merge. Hope we can move `rosdep` `Windows` support forward.

NOTE: The CI failure seems unrelated to the change and gets repro on `master` too.

seanyen:
@wjwwood Ping! This is ready to review and merge. Let me know any additional tests to run.

wjwwood:
My latest comment said it needed to be tested on Linux to ensure it works as it does without the patch. Has that been done?

seanyen:
> My latest comment said it needed to be tested on Linux to ensure it works as it does without the patch. Has that been done?

@wjwwood Not sure how extensive the matrix I should validate, but I ran it on containers (python3\debian10, python2.7\debain10, and python2.7/alpine3.10) and Ubuntu bionic WSL. I can run `rosdep init` and `rosdep update` and see it is working with this patch.

seanyen:
@tfoote Thanks for catching this. I updated the check.

nuclearsandwich:
This change passed CI when rebased as #749.

PR #657, opened 2019-01-29
Only release on platforms supported by rosdistro 0.7.0.

nuclearsandwich:
Closes a packaging regression on platforms no longer supported by
current rosdistro releases.

As of #655 (first released in 0.15.0) we're relying on a version of rosdistro which is not supported by older Ubuntu and Debian releases (ros-infrastructure/rosdistro#123). 

In addition to this change for future releases, the 0.15.0 release will be yanked from the platforms removed below to fix the regression.

### DIFF
--- a/stdeb.cfg
+++ b/stdeb.cfg
@@ -5,4 +5,4 @@ Conflicts: python3-rosdep, python-rosdep2, python3-rosdep2
 Conflicts3: python-rosdep, python-rosdep2, python3-rosdep2
 Copyright-File: LICENSE
 Suite: trusty utopic vivid wily xenial yakkety zesty artful bionic jessie stretch buster
-X-Python3-Version: >= 3.2
+X-Python3-Version: >= 3.4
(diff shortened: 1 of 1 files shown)


stdeb.cfg
```diff
@@ -4,5 +4,5 @@ Depends3: ca-certificates, python3-rospkg (>= 1.0.37), python3-yaml, python3-cat
 Conflicts: python3-rosdep, python-rosdep2, python3-rosdep2
 Conflicts3: python-rosdep, python-rosdep2, python3-rosdep2
 Copyright-File: LICENSE
-Suite: oneiric precise quantal raring saucy trusty utopic vivid wily xenial yakkety zesty artful bionic wheezy jessie stretch buster
+Suite: trusty utopic vivid wily xenial yakkety zesty artful bionic jessie stretch buster
 X-Python3-Version: >= 3.2
```
dirk-thomas: This should be bumped to 3.4.

dirk-thomas:
@k-okada FYI

stdeb.cfg
nuclearsandwich: It's still 3.2 in [rosdistro](https://github.com/ros-infrastructure/rosdistro/blob/a3052c9c2f9e790049da0755a4b5252cdf5ffc6a/stdeb.cfg#L15) and [catkin_pkg](https://github.com/ros-infrastructure/catkin_pkg/blob/f2a907a8f6e5e479e20065a3297d8718f510c69e/stdeb.cfg#L15) should we bump it there too?

dirk-thomas: PRs already pending: ros-infrastructure/catkin_pkg#249, ros-infrastructure/rospkg#161, ros-infrastructure/rosdistro#129

PR #661, opened 2019-02-15
[Windows] Add Azure CI file and fixing test breaks on Windows

seanyen:
This change is to add Azure DevOps pipelines for running Windows CI, and also fixes test breaks on Windows:

Highlight the changes on the production code side:
- `src/rosdep2/platforms/source.py`
  - add missing tarfile.close()
- `src/rosdep2/shell_utils.py`
  - add script suffix `.bat` on Windows to make the shell able to run the script.
- `src/rosdep2/main.py`
  - fix missing `os.EX_USAGE` on Windows causing runtime errors.
  - use portable `os.pathsep` to replace `:`
- `src/rosdep2/platforms/windows.py` & `src/rosdep2/__init__.py`
  - Add stub `rosdep2.platforms.windows` to avoid `Unsupported OS` error on Windows.
- `src/rosdep2/installers.py`
  - conditionally add 'sudo -H' based on availability of `os.geteuid()`.

Summary of changes on the test code side:
- Fix test cases to accommodate the `sudo -H` changes.
- Normalize `file://` URIs in test cases.
- Conditionally generate test script for bash scripts and batch scripts.
- Add missing mock function for `test_RosdepInstaller_install_resolved`
- Add alternative implementation for `os.path.samefile` for Python 2 on Windows.


seanyen-msft:
Azure CI result is published here: https://ros-win.visualstudio.com/ros-win/_build/results?buildId=3150

seanyen:
@nuclearsandwich Just a friendly ping. This is a PR focused on fixing test failures for `rosdep` on Windows. Let me know anything to iterate.

test/test_rosdep_source.py
```diff
@@ -56,6 +56,47 @@ def get_test_dir():
     return os.path.abspath(os.path.join(os.path.dirname(__file__), 'source'))
 
 
+def get_simple_script_exit_with(exitcode):
```
tfoote: This whole block of functions is duplicated across two test files and should be deduplicated.

test/test_rosdep_installers.py
```diff
@@ -614,7 +627,8 @@ def test_RosdepInstaller_install_resolved():
                 raise
             return True
     stdout_lines = [x.strip() for x in stdout.getvalue().split('\n') if x.strip()]
+    sudo_command = 'sudo -H ' if hasattr(os, 'geteuid') and os.geteuid() != 0 else ''
```
tfoote: Since we need to reimplement this logic in many places, I'd suggest that we make it into a utility function that we can test independently and then reuse throughout the implementations and tests.

tfoote (changes requested):
Overall this looks good to me. Though I saw that there were some places to reduce the size of the diff/additional code for better long term maintenance.

I'm a little surprised that the sudo commands were necessary to adjust for windows as in the unit tests I expect them to not be executed and just be string matching, but I think this is an improvement so we might as well use it.

seanyen:
@tfoote @nuclearsandwich I pushed new changes to address the duplicate logic feedbacks and rebase the code to the latest master. And the results are green for Windows CI: https://ros-win.visualstudio.com/ros-win/_build/results?buildId=4336

Hope you can take a look for the new iteration! Thanks!

seanyen:
@cottsay @tfoote @nuclearsandwich Please excuse me to ping you all. Hope this is still under the radar and it would be nice to see more Windows support for `rosdep`.

seanyen:
`travis` seemed to fail on a infrastructure issue:
```
Homebrew must be run under Ruby 2.6! You're running 2.3.3. 
```

nuclearsandwich:
@seanyen now that #751 is finally merged would you like to rework this PR to add Windows CI via GitHub Action?

Issue #641, opened 2018-11-14
Support for Package Sources


cottsay:
Fedora recently introduced a system they call 'Modularity' which is intended to address the too fast/too slow problem with packages.

This problem is a major pain point for running ROS on Fedora, because ROS system dependencies typically follow whatever versions are in the targeted Ubuntu releases at the time. It is very common for Fedora's package version to be different (and incompatible) with the versions used by Ubuntu, and therefore targeted by a particular ROS release. A great example is Gazebo, where Melodic targets Gazebo 9 and Fedora currently packages Gazebo 8.

Please feel free to read the Fedora Modular docs [[1](https://docs.fedoraproject.org/en-US/modularity/)], but to summarize the parts we care about here, a module (e.g. 'nodejs') has one or more streams (e.g. '8' or '12.x', or 'lts') and those streams contain differing versions of packages. When the module is enabled, packages from that module and selected stream are made available on the system.

There doesn't seem to be a precedent for this type of behavior in rosdep, but I think we could probably simplify the problem by thinking of modules as a package source of sorts. I would suggest two possible ways to implement this:
1. Treat package sources as packages themselves, and either:
    1. The required package can be installed implicitly as part of enabling the source
        ```yaml
        nodejs8:
          fedora:
            dnf-module: [nodejs:8]
        ```
    2. The package could leverage the existing `depends` support in rosdep, and the package and the source would then have separate rosdep keys. The module would only be enabled by the second key - no packages would be installed.
        ```yaml
        nodejs8:
          fedora:
            packages: [nodejs]
            depends: [nodejs8-module]
        nodejs8-module:
          fedora:
            dnf-module: [nodejs:8] 
        ```
2. A new concept (something like `PackageSourceInstaller`) could be added. This might look something like:
    ```yaml
    nodejs8:
      fedora:
        packages: [nodejs]
        sources:
          dnf-module: [nodejs:8]
    ```

I believe that (2) is the most clear, and could also be leveraged for many other opportunities, such as 3rd party repositories or Fedora COPR [[2](https://developer.fedoraproject.org/deployment/copr/about.html)].

I'm hoping for feedback on whether (2) would be desirable, or if I should just work around the problem using existing concepts as demonstrated by (1).

Thanks!

--scott

[1] https://docs.fedoraproject.org/en-US/modularity/
[2] https://developer.fedoraproject.org/deployment/copr/about.html

richmattes:
[...]

PR #667, opened 2019-03-11
Use only major version part of RHEL version number

cottsay:
At present, we don't have any rosdep rules split by release, so I don't believe this will change behavior for anyone.

We may need to differentiate between RHEL 7 and 8 because the Python 3 package names will likely change (from `python36-` to `python3-`). It would be pretty cumbersome to create rules for all of the minor versions of each RHEL release, so I'm proposing that we make rules based only on the major version.

For at least RHEL 6 and 7, the EPEL repository isn't broken up by minor release at all, so for packages sourced from there, the rule couldn't possibly differ anyway.

FYI @richmattes


richmattes:
Makes sense, and there's probably other deps that are different between 6 and 7 (and 8).

That said, in the Python case there's maybe a case to be made to embrace pip via the mechanism in #641.

PR #669, opened 2019-03-11
Use mockable OsDetect in get_os_version_type

cottsay:
Otherwise calling `get_os_name_and_version` may use the un-mocked instance.


cottsay:
I found a way to work around this issue. This approach doesn't work anyway.

PR #674, opened 2019-04-02
MXLinux support

gstavrinos:
Depends on ros-infrastructure/rospkg#167.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,7 +8,7 @@ setup(
     version=__version__,  # noqa:F821
     packages=['rosdep2', 'rosdep2.platforms'],
     package_dir={'': 'src'},
-    install_requires=['catkin_pkg >= 0.4.0', 'rospkg >= 1.0.37', 'rosdistro >= 0.7.0', 'PyYAML >= 3.1'],
+    install_requires=['catkin_pkg >= 0.4.0', 'rospkg >= 1.1.8', 'rosdistro >= 0.7.0', 'PyYAML >= 3.1'],
     test_suite='nose.collector',
     test_requires=['mock', 'nose >= 1.0'],
     scripts=['scripts/rosdep', 'scripts/rosdep-source'],
--- a/src/rosdep2/platforms/debian.py
+++ b/src/rosdep2/platforms/debian.py
@@ -31,7 +31,7 @@ from __future__ import print_function
 import subprocess
 import sys
 
-from rospkg.os_detect import OS_DEBIAN, OS_LINARO, OS_UBUNTU, OS_ELEMENTARY, OsDetect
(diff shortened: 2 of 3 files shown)


src/rosdep2/platforms/debian.py
```diff
@@ -83,6 +84,16 @@ def register_elementary(context):
...
 
 
+def register_mx(context):
+    # MX is an alias for Debian. If elementary is detected and it's
+    # not set as an override, force Debian.
+    (os_name, os_version) = context.get_os_name_and_version()
+    if os_name == OS_MX and not context.os_override:
+        print('rosdep detected OS: [%s] aliasing it to: [%s]' %
+              (OS_MX, OS_DEBIAN), file=sys.stderr)
+        context.set_os_override(OS_DEBIAN, context.os_detect.get_codename())
```
tfoote: From my testing, the codename for MX and Debian are different. You'll need to setup an appropriate aliaing method to remap the codenames through so that it knows which base debian packages are to be expected available.

gstavrinos: New (squashed) commit on ros-infrastructure/rospkg#167. It does not break compatibility with this one.

dirk-thomas:
This needs an updated version dependency on the next `rospkg` version since it uses new API introduced in ros-infrastructure/rospkg#167.

gstavrinos:
> This needs an updated version dependency on the next `rospkg` version since it uses new API introduced in [ros-infrastructure/rospkg#167](https://github.com/ros-infrastructure/rospkg/pull/167).

I guess 1.1.8 should be the next?

dirk-thomas:
> I guess 1.1.8 should be the next?

Correct :+1:

gstavrinos:
There you are! I will try to get the similar Elementary fix ready in the next days.

setup.py
```diff
@@ -8,7 +8,7 @@
     version=__version__,  # noqa:F821
     packages=['rosdep2', 'rosdep2.platforms'],
     package_dir={'': 'src'},
-    install_requires=['catkin_pkg >= 0.4.0', 'rospkg >= 1.0.37', 'rosdistro >= 0.7.0', 'PyYAML >= 3.1'],
+    install_requires=['catkin_pkg >= 0.4.0', 'rospkg >= 1.1.8', 'rosdistro >= 0.7.0', 'PyYAML >= 3.1'],
```
dirk-thomas: The version number needs to also be updated in these two lines: https://github.com/ros-infrastructure/rosdep/blob/fe63aaac265a2ef3823726250815c0d64e9e9b5c/stdeb.cfg#L2-L3

gstavrinos:
Fixed. Didn't know that. Sorry.

dirk-thomas:
`rospkg` 1.1.8 has been released.

gstavrinos:
I will make a new PR for Elementary OS as soon as the MX Linux changes are merged, so that I can fetch.

tfoote:
retriggered CI now that rospkg is out

PR #675, opened 2019-04-09
fix: make src install less verbose

daenny:
I have some custom .rdmanifest which install from source. Now the script is always printed in terminal (even with rosdep -q)
I think, this should be debug information.


tfoote:
The CI appears to be failing due to python hard deprecating yaml_load we worked around the externally accessible instances in #651 but left some internal ones.

daenny:
I added a commit which replaces the remaining `yaml.load()` with `yaml.safe_load()`

tfoote:
Thanks!

PR #680, opened 2019-05-13
Add dependency to sudo

yotabits:
Having sudo installed is mandatory to run rosdep properly.


yotabits:
> Although not technically necessary as it could be used as root. I think this is good to add as it is necessary for all of our recommended usages as non-root.

Also when deploying in docker it allows to not to have to install sudo manually prior to any rosdep call :)

PR #686, opened 2019-06-06
keys: add support for --ignore-src

### LINKED ISSUE #685: Apply --ignore-src to keys as well
I was recently surprised to learn that `-i`/`--ignore-packages-from-source`/`--ignore-src` only affects `check` and `install`, not `keys` (no error though, of course). That requires one to do their own filtering using e.g. `rospack list-names`, when the functionality obviously already exists in rosdep. How do the maintainers feel about supporting `-i` for `keys` as well, to filter out dependencies already satisfied by packages in the workspace? Happy to do the work.

kyrofa:
Currently `--ignore-src` is only supported by the `check` and `install` verbs. This PR resolves #685 by adding support in the `keys` verb as well, displaying dependencies that are not already satisfied by other packages in the workspace.

### DIFF
--- a/src/rosdep2/main.py
+++ b/src/rosdep2/main.py
@@ -61,7 +61,7 @@ from . import __version__
 from .core import RosdepInternalError, InstallFailed, UnsupportedOs, InvalidData, CachePermissionError, DownloadFailure
 from .installers import normalize_uninstalled_to_list
 from .installers import RosdepInstaller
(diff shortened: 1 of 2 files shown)


kyrofa:
@nuclearsandwich curious to get your thoughts on this.

kyrofa:
Hey @tfoote, any chance we can get this released? We're carrying some patches that we can drop once that happens.

PR #687, opened 2019-06-11
Adding changes for slow/bad networks

kunaltyagi:
On some networks, the following error is observed:
```python
Hit https://raw.githubusercontent.com/ros/rosdistro/master/rosdep/python.yaml
ERROR: unable to process source [https://raw.githubusercontent.com/ros/rosdistro/master/rosdep/ruby.yaml]:
        <urlopen error ('_ssl.c:711: The handshake operation timed out',)> (https://raw.githubusercontent.com/ros/rosdistro/master/rosdep/ruby.yaml)
```

This PR resolves the issue by trying the connection multiple times. Increasing timeout was tried but did not succeed.

The code can be made a bit more composable (if required) by converting the inner loops to functions.

### DIFF
--- a/src/rosdep2/main.py
+++ b/src/rosdep2/main.py
@@ -334,6 +334,10 @@ def _rosdep_main(args):
                       help="Affects the 'update' verb. "
                            'If specified end-of-life distros are being '
                            'fetched too.')
+    parser.add_option('--max-retries', dest='max_retries',
+                      default=5, type=int,
+                      help="Affects the 'update' verb. "
+                           'Trt to connect multiple number of times before signalling a failure')
 
     options, args = parser.parse_args(args)
     if options.print_version or options.print_all_versions:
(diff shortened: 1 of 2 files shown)


kunaltyagi:
Working on the errors

kunaltyagi:
I've captured IOError as well because somewhere an IOError raised by urllib it is not converted to DownloadError.

I've also observed that any errors for compliance with REP137 and REP143 are not reported. Currently, the code maintains it that way, though it doesn't make sense unless I'm missing something.

kunaltyagi:
I'm stuck on a mac unable to reproduce the errors. Please give me a little more time to resolve the errors.

KraftOreo:
In China it's kind of like a connection problem
You can go to /usr/lib/python2.7/dist-packages/rosdep2/  and change the DOWNLOAD_TIMEOUT = 15.0 to a bigger number in these three files. 
sources_list.py
gbpdistro_support.py
rep3.py

nuclearsandwich:
@kunaltyagi if you are able to resolve the merge conflicts here I can try to either assist with the test failures or help provide reproduction environments so you can continue work.

kunaltyagi:
@nuclearsandwich Repro env would be nice. Please give me a few days to clean this up

kunaltyagi:
I'll create a new PR, not from my master branch

kunaltyagi:
@nuclearsandwich I found that rosdistro already has a loader function which does pretty much the same thing I was trying to do. Instead of `urlopen`, we can just use `rosdistro.loader`.

The rosdistro also does a few additional stuff. If that makes no difference, we can easily have the reload + wait functionality in this pkg as well

PR #693, opened 2019-07-18
use `pip install --user` when `--as_root pip:false`, rather than `pip install`

k-okada:
Current implementation runs `pip install` command when `--as-root pip:false` is set.
In most cases, it does not work because it try to install the pip package under `/usr/local` without `sudo` command
So when we set `--as-root`, I think we should use `pip install --user` command.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -1,0 +1,6 @@
+installer-options:
+  pip:
+    additional-flags: user
+    sudo: false
+    command-options: --user
+
--- a/setup.py
+++ b/setup.py
@@ -12,6 +12,7 @@ setup(
     test_suite='nose.collector',
     test_requires=['mock', 'nose >= 1.0'],
     scripts=['scripts/rosdep', 'scripts/rosdep-source'],
+    data_files=[('etc/ros/rosdep', ['config/config.yaml'])],
     author='Tully Foote, Ken Conley',
     author_email='tfoote@osrfoundation.org',
     url='http://wiki.ros.org/rosdep',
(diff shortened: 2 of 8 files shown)


NikolausDemmel:
The original use case of adding this was on macos, where the system-wide install in /usr/local for brewed python shouldn't be done with root. So I don't think we should change the default.

Maybe instead some syntax like `--installer-args "pip:--user"` could be used.

k-okada:
thanks for feedback,  to be clear, "brewed python" means python on
osx/homebrew?
--
◉ Kei Okada



2019年7月18日(木) 19:23 Nikolaus Demmel <notifications@github.com>:

NikolausDemmel:
Yes, exactly, it means the python installed with homebrew (as opposed to the one shipped with macos). You can still do `pip install --user`, and maybe that is even a better idea if you are the only user for the ROS setup on that machine. I personally don't mind the change too much, but maybe some users rely on the current behaviour.

NikolausDemmel:
Oh look, there was already discussion on supporting `--user` back when `--as-root` was added, but nothing came of it in the end: https://github.com/ros-infrastructure/rosdep/pull/307#issuecomment-36345136

k-okada:
@NikolausDemmel thanks for comment, I have implemented `—installer-options` and add `rosdep/config.yaml` to support `—user`

nuclearsandwich:
@k-okada in addition to needing a rebase there are quite a few changes here which I think could be split out. I'm especially interested in the optparse -> argparse commits and I think the introduction of a config file to persist options will come with its own discussion and would benefit from a breakout PR as well. Would you be willing to open either or both of those?

PR #694, opened 2019-07-18
support version specific install for pip (and others)

k-okada:
We often have trouble on version problem of pip installed modules, and to avoid this it is very impotant to not to upgrade the python module version automatically (ofcourse do not use pip, is the best solution....)

This PR respect `version_eq` or other attribute in `depend` tag when running `pip install`

```
  <exec_depend version_lte="5.7">ipython-pip</exec_depend>
  <exec_depend version_eq="4.10">ipykernel-pip</exec_depend>  
```


NikolausDemmel:
I'm not maintaining rosdep, nor using it much these days, but just some thoughts from when we had discussed this in the past: 

- What happens if an incompatible version is already installed?
- What happens if you have different version statements (conflicting or compatible) for the same pip package in multiple ROS packages? 

You might argue to just leave all those decisions to pip, and that is probably a good idea if possible, but consider the following: I'm not 100% sure here, I believe rosdep tries to combine multiple packages to a single `pip install` command, if no other package managers appear in between in the list of resolutions. But what if they get split into separate invocations? So one case would be two different version requirements (compatible or not) for a pip package in the same call to `pip install`, and another case might be two different version requirements (compatible or not) for a pip package in two subsequent calls `pip install`. In the latter, might pip uninstall the package from the first invocation and make it silently incompatible?

Ideally, the intended behaviour in such cases would be documented and tested. But maybe it's also ok to keep it simple if it covers >95% of use cases. You should probably wait for some feedback on willingness/bandwidth to include this change from a maintainer.

130s:
Just raising +1 flag for adding this capability to utilize package.xml being able to define version limitations. We've ended up running`pip` manually for pkgs that need to be specific versions, which is obviously tedious and turned out to be prone to human errors. Concerns raised by @NikolausDemmel makes sense though.

knorth55:
+1 for this PR, but I can understand @NikolausDemmel 's concern.
But there will be the EOL of python2 soon, so we should move forward for this discussion because some python packages are modified to run only on python3 but carelessly released on python2.
In that case, I want to use version information for rosdep, too.

Timple:
> some python packages are modified to run only on python3 but carelessly released on python2.

In that case I think it would be okay to just pin the version in the rosdistro file because everybody on python2 will suffer from this.

src/rosdep2/installers.py
```diff
@@ -278,8 +278,9 @@ def get_depends(self, rosdep_args):
         """
         return []  # Default return empty list
 
-    def resolve(self, rosdep_args_dict):
+    def resolve(self, rosdep, rosdep_args_dict):
```
russkel, line 281: Can you please move this argument to after the `rosdep_args_dict` and add a default value for it so it doesn't unnecessarily break other projects that rely on `resolve`?

src/rosdep2/installers.py
```diff
@@ -336,7 +337,7 @@ def elevate_priv(self, cmd):
         """
         return (self.sudo_command.split() if self.as_root else []) + cmd
 
-    def resolve(self, rosdep_args):
+    def resolve(self, rosdep, rosdep_args):
```
russkel, line 340: Same here.

src/rosdep2/lookup.py
```diff
@@ -433,13 +438,13 @@ def resolve_all(self, resources, installer_context, implicit=False):
 
         return resolutions_flat, errors
 
-    def resolve(self, rosdep_key, resource_name, installer_context):
+    def resolve(self, rosdep, resource_name, installer_context):
```
russkel, line 441: Ditto.

russkel:
Alternatively, could you store the `Dependency` object in the `rosdep_args_dict` instead of changing the arguments of the `resolve` method?

src/rosdep2/installers.py
k-okada: @russkel thanks for comment, this makes much simpler, fixed the code

src/rosdep2/lookup.py
k-okada: this is not new argument, just to rename variable name, since we have changed from list of keys to `catkin_pkg.package.Dependency` object
``
        :param rosdep_key: rosdep key to resolve
        :param rosdeps: rosdep key (catkin_pkg.package.Dependency object) to resolve
``

russkel: Yep understood.

test/test_rosdep_lookup.py
```diff
@@ -343,26 +343,26 @@ def test_RosdepLookup_get_rosdeps():
         pass
 
     print(lookup.get_rosdeps('stack1_p1'))
-    assert set([d.name for d in lookup.get_rosdeps('stack1_p1')]) == set(['stack1_dep1', 'stack1_p1_dep1', 'stack1_p1_dep2'])
-    assert set([d.name for d in lookup.get_rosdeps('stack1_p1', implicit=False)]) == set(['stack1_dep1', 'stack1_p1_dep1', 'stack1_p1_dep2'])
+    assert set({d.name for d in lookup.get_rosdeps('stack1_p1')}) == set(['stack1_dep1', 'stack1_p1_dep1', 'stack1_p1_dep2'])
```
russkel, line 346: Set comprehension doesn't need the outer `set()`

JWhitleyWork:
@k-okada Would you mind taking a look at my new draft #901 which covers a small portion of what you've done here but on the install-verification side? This at least allows you to specify `pip` entries in YAML files which have a version requirement and verify that they were installed with that specific version.

PR #697, opened 2019-08-04
Add support for Nix/NixOS

lopsided98:
This is part of my effort to add support for [Nix](https://nixos.org/) to ROS. It depends on https://github.com/ros-infrastructure/rospkg/pull/171.

The Nix package manager is often used with NixOS, but it runs on any Linux distribution as well as macOS. In order to fit cleanly into the ROS infrastructure, I am pretending that Nix and NixOS have the traditional package manager and distribution relationship. This means that the rosdep OS key for Nix is `nixos`, even though this key may be used on multiple distributions. When packages are built, the build sandbox prevents OS detection from working anyway, so the `ROS_OS_OVERRIDE` environment variable is used to force the correct value.

`rosdep` will not be particularly useful to end users, because its imperative package management style does not fit Nix very well. Instead, it will be used to resolve dependencies when generating packages using `superflore`.

I am working on adding support for Nix to various ROS projects:
* https://github.com/lopsided98/rospkg/tree/nixos-support
* https://github.com/lopsided98/rosdep/tree/nixos-support
* https://github.com/lopsided98/superflore/tree/nixos-support

I maintain a [Superflore generated nixpkgs overlay](https://github.com/lopsided98/nix-ros-overlay) and have successfully built over 3600 ROS packages on a [Hydra](https://nixos.org/hydra/) build server (not yet publicly accessible)

I also made a small cleanup to the Arch Linux platform.

### DIFF
--- a/src/rosdep2/__init__.py
+++ b/src/rosdep2/__init__.py
@@ -59,8 +59,9 @@ def create_default_installer_context(verbose=False):
     from .platforms import arch
     from .platforms import cygwin
     from .platforms import debian
-    from .platforms import openembedded
     from .platforms import gentoo
+    from .platforms import nix
+    from .platforms import openembedded
     from .platforms import opensuse
     from .platforms import osx
     from .platforms import pip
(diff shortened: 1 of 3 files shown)


dirk-thomas:
> it won't pass CI until the rospkg update is merged and released first.

`rospkg` 1.1.10 has been released with the referenced change.

ros-discourse:

https://discourse.ros.org/t/build-systems-package-management-and-nix/41488/16

PR #695, opened 2019-07-18
fixing unguarded next generator

danielwangksu:
fix https://github.com/ros-infrastructure/rosdep/issues/691
Signed-off-by: Daniel Wang <daniel.wang@canonical.com>

PR #701, opened 2019-08-27
Guard next(inter)

### LINKED ISSUE #691: Unguarded next in generator
if `lines` are exhausted will raise StopIteration exception
https://github.com/ros-infrastructure/rosdep/blob/ab1f06ed84ee1e716eb611457b1168710415ea7a/src/rosdep2/platforms/debian.py#L143-L157

artivis:
Catch possible `StopIteration` exception on `next(iter)` call.

This PR replaces #695 which is not passing CI.
Fix #691

Signed-off-by: artivis <jeremie.deray@canonical.com>

### DIFF
--- a/src/rosdep2/platforms/debian.py
+++ b/src/rosdep2/platforms/debian.py
@@ -140,20 +140,18 @@ def _read_apt_cache_showpkg(packages, exec_fn=None):
 
         header = 'Package: %s' % p
         # proceed to Package header
-        while True:
-            try:
-                if next(lines) == header:
-                    break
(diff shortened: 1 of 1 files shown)


artivis:
Travis is happy but Codecov is not. How much of a big deal is it given the lost delta?

nuclearsandwich:
@artivis I added 50990e33 after a discussion with @cottsay (his idea to fall through, all I did was actually type it up and commit it) what do you both think?

artivis:
I guess that works too, and it is more compact. +1

PR #703, opened 2019-09-04
tests: don't assume euid != 0

### LINKED ISSUE #702: Tests assume running as non-root, while code supports it
There are a few tests with `sudo` hard-coded (e.g. [this one](https://github.com/ros-infrastructure/rosdep/blob/master/test/test_rosdep_opensuse.py#L51)), despite the code having the logic to determine if sudo is needed. This makes the tests fail when running in an environment where `sudo` is not required. Ideally the tests would not be as dependent on the running environment.

kyrofa:
If tests rely on the environment being a certain way, they should ensure it meets expectations. This PR fixes #702 by updating the tests to ensure they're testing multiple euid conditions, and setting it explicitly where required.

### DIFF
--- a/test/test_rosdep_alpine.py
+++ b/test/test_rosdep_alpine.py
@@ -65,27 +65,30 @@ def test_ApkInstaller():
     from rosdep2.platforms.alpine import ApkInstaller
 
     @patch.object(ApkInstaller, 'get_packages_to_install')
-    def test(mock_method):
+    def test(expected_prefix, mock_method):
         installer = ApkInstaller()
         mock_method.return_value = []
         assert [] == installer.get_install_command(['nonexistingfakepackage'])
 
         mock_method.return_value = ['a-dev', 'b-dev']
 
-        expected = [['sudo', '-H', 'apk', 'add', 'a-dev', 'b-dev']]
+        expected = [expected_prefix + ['apk', 'add', 'a-dev', 'b-dev']]
         val = installer.get_install_command(['notused'], interactive=False, quiet=False)
         assert val == expected, 'Result was: %s' % val
 
-        expected = [['sudo', '-H', 'apk', 'add', '--interactive', 'a-dev', 'b-dev']]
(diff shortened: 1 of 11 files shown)


kyrofa:
@nuclearsandwich, @cottsay look sane to you?

kyrofa:
> Is there a reason that the expected_prefix must be the first argument

Just because the patch decorators need to go at the end.

> [...] and why it doesn't have a default value?

Both for clarity to the caller as well as so we didn't need to use a `None` canary value and have logic in each test to initialize to an empty list (as a default value of `[]` is a bad idea).

kyrofa:
@nuclearsandwich can I do anything to help this along?

kyrofa:
@nuclearsandwich any further thoughts?

nuclearsandwich:
I've come close to this so many times. Since something is better than nothing, I keep getting hung up on how duplicative this is but I don't have anything better. I need to do a full pass through rosdep PRs. The point release today was to fix an outstanding bug and rather than let it keep longer I just took that.

kyrofa:
> I keep getting hung up on how duplicative this is but I don't have anything better

Haha, I hear you. Without a restructuring of the existing tests I'm not sure this can be significantly improved. That said, if you prefer we can change it to only test running as root/non-root once and then mock it out everywhere else to use sudo, which would be a one-line change in the other tests. That's not quite as regression-proof, but the diif would be smaller.

Arnatious:
@nuclearsandwich @kyrofa has https://github.com/ros-infrastructure/rosdep/pull/680 invalidated this PR?

kyrofa:
@Arnatious no, [it doesn't](https://github.com/ros-infrastructure/rosdep/blob/master/src/rosdep2/installers.py#L328).

nuclearsandwich:
Rebased this against master in order to resolve the merge conflict.

kyrofa:
Thanks @nuclearsandwich!

PR #706, opened 2019-09-19
Update shebang to python3

ruffsl:
It seems the debian install is broken without this, e.g:

```
$ rosdep
  File "/usr/bin/rosdep", line 3, in <module>
    from rosdep2.main import rosdep_main
ImportError: No module named rosdep2.main
```

### DIFF
--- a/scripts/rosdep
+++ b/scripts/rosdep
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 from rosdep2.main import rosdep_main
 rosdep_main()


ruffsl:
Ok, never mind this PR then.

nuclearsandwich:
I've not substantiated the theory but my hypothesis is that the incorrect line endings present when creating the 0.16.0 release interfered with the shebang modification during release. 0.16.1 which was generated from a fixed repository clone appears to work on the bootstrap repository and is moving out to the main repositories now.

PR #708, opened 2019-09-25
Set ROS_PYTHON_VERSION if unset

sloretz:
Requires ros-infrastructure/rosdistro#145

This sets `ROS_PYTHON_VERSION` if it is unset. This enables rosdep to resolve conditional dependencies using `ROS_PYTHON_VERSION` before building a ROS distribution from source.

If `--rosdistro` is given, it tries to use the `python_version` field from the rosdistro index. Otherwise, it uses the same version of python used to invoke rosdep as mentioned in ros-infrastructure/rep#201

To try out the `python_version` attribute, update using a custom rosdistro index url

```
ROSDISTRO_INDEX_URL=https://raw.githubusercontent.com/ros/rosdistro/dirk-thomas/add-python_version-to-index/index-v4.yaml rosdep update
```

To test, get a package using conditional dependencies and run the commands below with no ROS workspace sourced.

```
git clone https://github.com/ros/urdf_parser_py.git -b py3-port
```

Before this PR
```
(env3) sloretz@7895d0788081:/tmp$ rosdep check --from-paths urdf_parser_py/ --ignore-src --skip-keys="rospy catkin"
All system dependencies have been satisfied
```

no rosdistro after this PR
```
sloretz@b0e9deb48a5e:~$ rosdep check --from-paths urdf_parser_py/ --ignore-src --skip-keys="rospy catkin"
WARNING: ROS_PYTHON_VERSION is unset. Defaulting to 3
System dependencies have not been satisfied:
apt	python3-lxml
apt	python3-mock
```

melodic
```
sloretz@b0e9deb48a5e:~$ rosdep check --rosdistro melodic --from-paths urdf_parser_py/ --ignore-src --skip-keys="rospy catkin"
System dependencies have not been satisfied:
apt	python-lxml
apt	python-yaml
apt	python-mock
```

noetic
```
sloretz@b0e9deb48a5e:~$ rosdep check --rosdistro noetic --from-paths urdf_parser_py/ --ignore-src --skip-keys="rospy catkin"
System dependencies have not been satisfied:
apt	python3-lxml
apt	python3-mock
```

### DIFF
--- a/src/rosdep2/main.py
+++ b/src/rosdep2/main.py
@@ -62,6 +62,7 @@ from .core import RosdepInternalError, InstallFailed, UnsupportedOs, InvalidData
 from .installers import normalize_uninstalled_to_list
 from .installers import RosdepInstaller
 from .lookup import RosdepLookup, ResolutionError, prune_catkin_packages
+from .meta import MetaDatabase
 from .rospkg_loader import DEFAULT_VIEW_KEY
 from .sources_list import update_sources_list, get_sources_cache_dir,\
     download_default_sources_list, SourcesListLoader, CACHE_INDEX,\
@@ -384,11 +385,6 @@ def _rosdep_main(args):
     if options.ros_distro:
         os.environ['ROS_DISTRO'] = options.ros_distro
 
-    if 'ROS_PYTHON_VERSION' not in os.environ:
-        print('WARNING: ROS_PYTHON_VERSION is unset. Defaulting to {}'.format(sys.version[0]), file=sys.stderr)
-        # Default to same python version used to invoke rosdep
-        os.environ['ROS_PYTHON_VERSION'] = sys.version[0]
-
     # Convert list of keys to dictionary
     options.as_root = dict((k, str_to_bool(v)) for k, v in key_list_to_dict(options.as_root).items())
 
@@ -396,6 +392,18 @@ def _rosdep_main(args):
         check_for_sources_list_init(options.sources_cache_dir)
     elif command not in ['fix-permissions']:
         setup_proxy_opener()
+
+    if 'ROS_PYTHON_VERSION' not in os.environ and options.ros_distro:
+        # Set python version to version used by ROS distro
+        python_versions = MetaDatabase().get('ROS_PYTHON_VERSION')
+        if options.ros_distro in python_versions:
+            os.environ['ROS_PYTHON_VERSION'] = str(python_versions[options.ros_distro])
+
+    if 'ROS_PYTHON_VERSION' not in os.environ:
+        # Default to same python version used to invoke rosdep
+        print('WARNING: ROS_PYTHON_VERSION is unset. Defaulting to {}'.format(sys.version[0]), file=sys.stderr)
+        os.environ['ROS_PYTHON_VERSION'] = sys.version[0]
+
     if command in _command_rosdep_args:
         return _rosdep_args_handler(command, parser, options, args)
     elif command in _command_no_args:
--- a/src/rosdep2/main.py
+++ b/src/rosdep2/main.py
@@ -384,6 +384,11 @@ def _rosdep_main(args):
     if options.ros_distro:
         os.environ['ROS_DISTRO'] = options.ros_distro
 
+    if 'ROS_PYTHON_VERSION' not in os.environ:
+        print('WARNING: ROS_PYTHON_VERSION is unset. Defaulting to {}'.format(sys.version[0]), file=sys.stderr)
+        # Default to same python version used to invoke rosdep
+        os.environ['ROS_PYTHON_VERSION'] = sys.version[0]
+
     # Convert list of keys to dictionary
     options.as_root = dict((k, str_to_bool(v)) for k, v in key_list_to_dict(options.as_root).items())
 
--- a/src/rosdep2/main.py
+++ b/src/rosdep2/main.py
@@ -383,6 +383,13 @@ def _rosdep_main(args):
     args = args[1:]
 
     if options.ros_distro:
+        if 'ROS_DISTRO' in os.environ and os.environ['ROS_DISTRO'] != options.ros_distro:
+            # user has a different workspace sourced, use --rosdistro
+            print('WARNING: given --rosdistro {} but ROS_DISTRO is "{}". Ignoring environment.'.format(
+                options.ros_distro, os.environ['ROS_DISTRO']))
+            # Use python version from --rosdistro
+            if 'ROS_PYTHON_VERSION' in os.environ:
+                del os.environ['ROS_PYTHON_VERSION']
         os.environ['ROS_DISTRO'] = options.ros_distro
 
     # Convert list of keys to dictionary
@@ -393,11 +400,11 @@ def _rosdep_main(args):
     elif command not in ['fix-permissions']:
         setup_proxy_opener()
 
-    if 'ROS_PYTHON_VERSION' not in os.environ and options.ros_distro:
+    if 'ROS_PYTHON_VERSION' not in os.environ and 'ROS_DISTRO' in os.environ:
         # Set python version to version used by ROS distro
         python_versions = MetaDatabase().get('ROS_PYTHON_VERSION')
-        if options.ros_distro in python_versions:
-            os.environ['ROS_PYTHON_VERSION'] = str(python_versions[options.ros_distro])
+        if os.environ['ROS_DISTRO'] in python_versions:
+            os.environ['ROS_PYTHON_VERSION'] = str(python_versions[os.environ['ROS_DISTRO']])
 
     if 'ROS_PYTHON_VERSION' not in os.environ:
         # Default to same python version used to invoke rosdep
--- a/src/rosdep2/sources_list.py
+++ b/src/rosdep2/sources_list.py
@@ -46,6 +46,7 @@ except ImportError:
 from .cache_tools import compute_filename_hash, PICKLE_CACHE_EXT, write_atomic, write_cache_file
 from .core import InvalidData, DownloadFailure, CachePermissionError
 from .gbpdistro_support import get_gbprepo_as_rosdep_data, download_gbpdistro_as_rosdep_data
+from .meta import MetaDatabase
 
 try:
     import urlparse
@@ -483,6 +484,8 @@ def update_sources_list(sources_list_dir=None, sources_cache_dir=None,
 
     # Additional sources for ros distros
     # In compliance with REP137 and REP143
+    python_versions = {}
+
     print('Query rosdistro index %s' % get_index_url())
     for dist_name in sorted(get_index().distributions.keys()):
         distribution = get_index().distributions[dist_name]
@@ -493,11 +496,17 @@ def update_sources_list(sources_list_dir=None, sources_cache_dir=None,
         print('Add distro "%s"' % dist_name)
         rds = RosDistroSource(dist_name)
         rosdep_data = get_gbprepo_as_rosdep_data(dist_name)
+        # Store Python version from REP153
+        if distribution.get('python_version'):
+            python_versions[dist_name] = distribution.get('python_version')
         # dist_files can either be a string (single filename) or a list (list of filenames)
         dist_files = distribution['distribution']
         key = _generate_key_from_urls(dist_files)
         retval.append((rds, write_cache_file(sources_cache_dir, key, rosdep_data)))
         sources.append(rds)
+
+    # cache metadata that isn't a source list
+    MetaDatabase().set('ROS_PYTHON_VERSION', python_versions)
 
     # Create a combined index of *all* the sources.  We do all the
     # sources regardless of failures because a cache from a previous
--- a/src/rosdep2/sources_list.py
+++ b/src/rosdep2/sources_list.py
@@ -31,9 +31,7 @@ from __future__ import print_function
 
 import os
 import sys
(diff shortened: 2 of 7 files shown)


dirk-thomas:
With ros-infrastructure/rep#207 in flight that information could be used here too.

sloretz:
@nuclearsandwich added warning in 88816ef

sloretz:
> This should use the information from the updated REP 153.

REP 153 `python_version` can be used instead of `sys.version[0]` when `--rosdistro` is set. That seems more like an enhancement than a blocker.

dirk-thomas:
> REP 153 `python_version` can be used instead of `sys.version[0]` when `--rosdistro` is set. That seems more like an enhancement than a blocker.

> it has nearly a 50-50 chance of being the wrong choice.

With the high probability of doing the wrong thing I wouldn't call this an enhancement. I think we should do it right when we can (when `--rosdistro` is available).

sloretz:
> I think we should do it right when we can (when --rosdistro is available).

PR and PR description updated

src/rosdep2/main.py
```diff
@@ -391,6 +392,18 @@ def _rosdep_main(args):
         check_for_sources_list_init(options.sources_cache_dir)
     elif command not in ['fix-permissions']:
         setup_proxy_opener()
+
+    if 'ROS_PYTHON_VERSION' not in os.environ and options.ros_distro:
```
dirk-thomas: If the ROS distro has information about the Python version I would expect it to be used and override the one set in the environment. Otherwise releasing a package for Noetic when Melodic is sourced will result in incorrect Debian control files.

wjwwood: Shouldn't `options.ros_distro` take precedence over the environment variable?

src/rosdep2/sources_list.py
```diff
@@ -493,12 +496,18 @@ def update_sources_list(sources_list_dir=None, sources_cache_dir=None,
         print('Add distro "%s"' % dist_name)
         rds = RosDistroSource(dist_name)
         rosdep_data = get_gbprepo_as_rosdep_data(dist_name)
+        # Store Python version from REP153
+        if distribution.get('python_version'):
+            python_versions[dist_name] = distribution.get('python_version')
```
dirk-thomas: For this to be available `rosdep` should declare a version dependency on the upcoming `rosdistro` version 0.7.5.

src/rosdep2/main.py
sloretz: `options.ros_distro` overriding `ROS_PYTHON_VERSION` might be a problem for Arch Linux users. They seem to be using Python 3: https://aur.archlinux.org/packages/ros-melodic-catkin/ with melodic. If `options.ros_distro` takes precedence then with #709 pip dependencies would be installed with Python 2 using Melodic on Arch.

dirk-thomas: See https://github.com/ros-infrastructure/rosdep/pull/708#discussion_r331204564 why on other platforms the rosdistro configured value must have precedence.

sloretz: I think this PR only affects the CLI tool. Bloom still needs to pick the right Python version when generating Debian control files.

dirk-thomas: So what happens when I have Melodic sourced but created a Noetic workspace and trying to install its dependencies using `rosdep` and explicitly set `--rosdistro noetic`?

sloretz: > So what happens when I have Melodic sourced but created a Noetic workspace and trying to install its dependencies using rosdep and explicitly set --rosdistro noetic?

All the bad things one would expect. If the melodic workspace has the `ros_environment` package, the environment variable `ROS_PYTHON_VERSION` would be set to `2`. Conditional dependencies in the Noetic workspace would be evaluated using `ROS_PYTHON_VERSION=2`, which is opposite of the intent of the package authors in the Noetic workspace. With #709 pip dependencies would be installed with Python 2 as well. This could be a problem for users because it's easy to accidentally have a `melodic` workspace sourced (like, if it's been added to `~/.bashrc`).

 Is there something we can do to address this case without breaking platforms that jumped to Python 3 already? Keeping with the same scenario of a user having a Melodic workspace sourced, say they manage to install all the right dependencies of their new Noetic workspace, and now they're ready to run `catkin_make`. What does `catkin_make` do to prevent packages in the `Noetic` workspace from using package in the `Melodic` workspace. Maybe we can do the same thing here?

src/rosdep2/sources_list.py
sloretz: Bumped dependency on `rosdistro` to `0.7.5` in f32222f

src/rosdep2/main.py
dirk-thomas: > What does `catkin_make` do to prevent packages in the `Noetic` workspace from using package in the `Melodic` workspace?

Nothing. For `catkin` the already sourced workspace is just an underlay which it will happily fall back to.

Only the `ros_environment` package does some sanity checking and warning the user if they mix ROS distros: see https://github.com/ros/ros_environment/blob/7ede5949279af45582aefd124d6d69a9125d985e/env-hooks/1.ros_distro.sh.in#L4

On a CMake level `catkin` will simply use the `ROS_PYTHON_VERSION` env var (see https://github.com/ros/catkin/blob/77822d31cd84c052e776f58c4d5f5630e0760725/cmake/python.cmake#L3) - in this case Python 2 for Noetic :boom:

sloretz:
> I haven't looked at the added MetaDatabase logic. I will leave that up to the maintainer(s) of rosdep.

Added tests in 830db970857c88192db5c904def2150d97751ef1. They pass in Python 2 and Python 3 as of 8a2f9542a6b2f22150dcac0d8e8863c564a84baa.

src/rosdep2/main.py
sloretz: I think we can do both. Thoughts on dd4bf62? If `--rosdistro` and `ROS_DISTRO` differ, then rosdep prefers `--rosdistro` and the index over the environment. If `--rosdistro` is unspecified or the same as `ROS_DISTRO`, then rosdep prefers the environment over the index.

In the case of a user with a melodic workspace sourced trying to get dependencies for a noetic workspace, rosdep will use Python 3 since that's in the index for noetic.

```
(rosdep-env3) sloretz@nxt /t/urdf_parser_py> env ROS_DISTRO=melodic ROS_PYTHON_VERSION=2 rosdep check --from-paths . --skip-keys="rospy catkin" --rosdistro noetic
WARNING: given --rosdistro noetic but ROS_DISTRO is "melodic". Ignoring environment.
System dependencies have not been satisfied:
apt	python3-lxml
apt	python3-mock
```

In the case of a user on an unsupported platform using melodic and Python 3 to get dependencies for a melodic workspace, rosdep will use Python 3 since that's the version in the environment variable.


```
(rosdep-env3) sloretz@nxt /t/urdf_parser_py> env ROS_DISTRO=melodic ROS_PYTHON_VERSION=3 rosdep check --from-paths . --skip-keys="rospy catkin"
System dependencies have not been satisfied:
apt	python3-lxml
apt	python3-mock
(rosdep-env3) sloretz@nxt /t/urdf_parser_py> env ROS_DISTRO=melodic ROS_PYTHON_VERSION=3 rosdep check --from-paths . --skip-keys="rospy catkin" --rosdistro melodic
System dependencies have not been satisfied:
apt	python3-lxml
apt	python3-mock
```

And checking normal cases, melodic sourced fetching dependencies for a melodic workspace always uses python 2
```
(rosdep-env3) sloretz@nxt /t/urdf_parser_py> env ROS_DISTRO=melodic ROS_PYTHON_VERSION=2 rosdep check --from-paths . --skip-keys="rospy catkin"
All system dependencies have been satisfied
(rosdep-env3) sloretz@nxt /t/urdf_parser_py> env ROS_DISTRO=melodic ROS_PYTHON_VERSION=2 rosdep check --from-paths . --skip-keys="rospy catkin" --rosdistro melodic
All system dependencies have been satisfied
```

noetic sourced fetching dependencies for a noetic workspace always uses python 3
```
(rosdep-env3) sloretz@nxt /t/urdf_parser_py> env ROS_DISTRO=noetic ROS_PYTHON_VERSION=3 rosdep check --from-paths . --skip-keys="rospy catkin"
System dependencies have not been satisfied:
apt	python3-lxml
apt	python3-mock
(rosdep-env3) sloretz@nxt /t/urdf_parser_py> env ROS_DISTRO=noetic ROS_PYTHON_VERSION=3 rosdep check --from-paths . --skip-keys="rospy catkin" --rosdistro noetic
System dependencies have not been satisfied:
apt	python3-lxml
apt	python3-mock
```

No workspace sourced always uses `ROS_PYTHON_VERSION`

```
(rosdep-env3) sloretz@nxt /t/urdf_parser_py> env ROS_PYTHON_VERSION=3 rosdep check --from-paths . --skip-keys="rospy catkin" --rosdistro noetic
System dependencies have not been satisfied:
apt	python3-lxml
apt	python3-mock
(rosdep-env3) sloretz@nxt /t/urdf_parser_py> env ROS_PYTHON_VERSION=2 rosdep check --from-paths . --skip-keys="rospy catkin" --rosdistro noetic
All system dependencies have been satisfied
```

dirk-thomas: Sounds like a very good heuristic to me :+1:

sloretz:
CI failure looks like a temporary network issue

```
======================================================================

ERROR: test.test_rosdep_source.test_install_source

----------------------------------------------------------------------

  File "/opt/python/3.6.7/lib/python3.6/urllib/request.py", line 1318, in do_open

    encode_chunked=req.has_header('Transfer-encoding'))

  File "/opt/python/3.6.7/lib/python3.6/http/client.py", line 1239, in request

    self._send_request(method, url, body, headers, encode_chunked)

  File "/opt/python/3.6.7/lib/python3.6/http/client.py", line 1285, in _send_request

    self.endheaders(body, encode_chunked=encode_chunked)

  File "/opt/python/3.6.7/lib/python3.6/http/client.py", line 1234, in endheaders

    self._send_output(message_body, encode_chunked=encode_chunked)

  File "/opt/python/3.6.7/lib/python3.6/http/client.py", line 1026, in _send_output

    self.send(msg)

  File "/opt/python/3.6.7/lib/python3.6/http/client.py", line 964, in send

    self.connect()

  File "/opt/python/3.6.7/lib/python3.6/http/client.py", line 1392, in connect

    super().connect()

  File "/opt/python/3.6.7/lib/python3.6/http/client.py", line 936, in connect

    (self.host,self.port), self.timeout, self.source_address)

  File "/opt/python/3.6.7/lib/python3.6/socket.py", line 724, in create_connection

    raise err

  File "/opt/python/3.6.7/lib/python3.6/socket.py", line 713, in create_connection

    sock.connect(sa)

TimeoutError: [Errno 110] Connection timed out

During handling of the above exception, another exception occurred:

  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/nose/case.py", line 198, in runTest

    self.test(*self.arg)

  File "/home/travis/build/ros-infrastructure/rosdep/test/test_rosdep_source.py", line 318, in test_install_source

    install_source(resolved)

  File "/home/travis/build/ros-infrastructure/rosdep/src/rosdep2/platforms/source.py", line 278, in install_source

    f = urlretrieve(resolved.tarball, filename)

  File "/opt/python/3.6.7/lib/python3.6/urllib/request.py", line 248, in urlretrieve

    with contextlib.closing(urlopen(url, data)) as fp:

  File "/opt/python/3.6.7/lib/python3.6/urllib/request.py", line 223, in urlopen

    return opener.open(url, data, timeout)

  File "/opt/python/3.6.7/lib/python3.6/urllib/request.py", line 526, in open

    response = self._open(req, data)

  File "/opt/python/3.6.7/lib/python3.6/urllib/request.py", line 544, in _open

    '_open', req)

  File "/opt/python/3.6.7/lib/python3.6/urllib/request.py", line 504, in _call_chain

    result = func(*args)

  File "/opt/python/3.6.7/lib/python3.6/urllib/request.py", line 1361, in https_open

    context=self._context, check_hostname=self._check_hostname)

  File "/opt/python/3.6.7/lib/python3.6/urllib/request.py", line 1320, in do_open

    raise URLError(err)

urllib.error.URLError: <urlopen error [Errno 110] Connection timed out>
```

src/rosdep2/main.py
sloretz: @wjwwood I think this is resolved by dd4bf62 (explanation in https://github.com/ros-infrastructure/rosdep/pull/708#discussion_r331204564). Would you mind having another look at this PR?

src/rosdep2/main.py
```diff
@@ -391,6 +399,18 @@ def _rosdep_main(args):
         check_for_sources_list_init(options.sources_cache_dir)
     elif command not in ['fix-permissions']:
         setup_proxy_opener()
+
+    if 'ROS_PYTHON_VERSION' not in os.environ and 'ROS_DISTRO' in os.environ:
+        # Set python version to version used by ROS distro
+        python_versions = MetaDatabase().get('ROS_PYTHON_VERSION')
```
acarrillo: Looks like `MetaDatabase().get('ROS_PYTHON_VERSION')` can return `None` here, it's causing my CI builds to fail after they upgraded to rosdep 0.16.2: https://github.com/ros-infrastructure/rosdep/issues/720

sloretz: Fix proposed in #721

PR #727, opened 2019-10-28
Selection of dependency types with cmd line option

nalt:
Select dependency types (build, buildtool, run, test) to install with --dependency-types. can be given multiple. Standard behaviour (all types) unchanged.

In rospack.py, a fake and empty options object was created for the call to _get_default_RosdepLookup. This was replaced by a similarly hacky global object for command line options.


nuclearsandwich:
Question for other rosdep contributors and maintainer: An approach similar to this one was proposed in https://github.com/ros-infrastructure/rosdep/pull/646#issuecomment-490032583 with a slightly different interface. Does anyone have a preference between the two and if so for which one and why?

gavanderhoorn:
@nuclearsandwich: you are essentially asking whether ppl would have a preference for:

```
--dependency-types <names_of_dep_types_separated_by_spaces>
```

or:

```
--dependency-types <name_of_dep_type> --dependency-types <name_of_dep_type> etc
```

right?

NikolausDemmel:
> @nuclearsandwich: you are essentially asking whether ppl would have a preference for:
> 
> ```
> --dependency-types <names_of_dep_types_separated_by_spaces>
> ```
> 
> or:
> 
> ```
> --dependency-types <name_of_dep_type> --dependency-types <name_of_dep_type> etc
> ```
> 
> right?

Side remark: On this question there is precedent for offering both options at the same time in rosdep: https://github.com/ros-infrastructure/rosdep/blob/c1c220ee671f47ed6e760e115f493385f9b27624/src/rosdep2/main.py#L404-L406

ruffsl:
IMO, inputs separated by spaces are bit friendlier to work with in scripts and pipes using the CLI.

dirk-thomas:
> inputs separated by spaces are bit friendlier to work with

Btw this is also the only way supported by `argparse`. So in order to not do custom command line argument parsing I would say we should use what is offered by the "standard" in Python.

mortenfyhn:
Just chiming in that I'm very excited about this feature.

ros-discourse:

https://discourse.ros.org/t/ros-noetic-ninjemys-release/14262/9

iluetkeb:
This seems to have stalled a bit -- any particular reason, or can we revive it?

mateus-amarante:
I opened a Pull Request to @nalt's branch (https://github.com/nalt/rosdep/pull/1/) finishing this update. Let's see if he answers. I'm not sure this is the best move in this situation. Alternatively, I could open a new PR but then we would have two PR's for the same thing. Please, let me know if there is a more adequate action.

gavanderhoorn:
Getting this merged would make using multi-stage Docker builds finally really scalable (no more manual runtime dependency management :tada:).

What's the major blocker?

nuclearsandwich:
> What's the major blocker?

At this point it's review time. I'll schedule it for next week and try my best to meet that appointment.

Makefile
```diff
@@ -27,4 +27,4 @@ testsetup:
 	echo "running rosdep tests"
 
 test: testsetup
-	nosetests --with-coverage --cover-package=rosdep2 --with-xunit test
+	nosetests3 --with-coverage --cover-package=rosdep2 --with-xunit test
```
mateus-amarante, line 30:
```suggestion
	nosetests --with-coverage --cover-package=rosdep2 --with-xunit test
```
It was my bad. I accidentally committed it.

src/rosdep2/main.py
```diff
@@ -295,6 +300,7 @@ def setup_environment_variables(ros_distro):
 
 def _rosdep_main(args):
     # sources cache dir is our local database.
+    global _global_options
```
mateus-amarante, line 303:
```suggestion
```
This is not needed

src/rosdep2/main.py
```diff
@@ -368,8 +374,14 @@ def _rosdep_main(args):
...
                            'If specified end-of-life distros are being '
                            'fetched too.')
+    parser.add_option('-t', '--dependency-types', dest='dependency_types',
+                      type="choice", choices=("build", "buildtool", "build_export", "exec", "run", "test", "doc"),
+                      default=[], action='append',
+                      help='Dependency types to install, can be given multiple times. '
+                           'Chose from build, buildtool, build_export, exec, run, test, doc. Default: all except doc.')
 
     options, args = parser.parse_args(args)
+    _global_options = options
```
mateus-amarante, line 384:
```suggestion
```
Not needed

src/rosdep2/lookup.py
```diff
@@ -339,6 +339,8 @@ def create_from_rospkg(rospack=None, rosstack=None,
           instance used to crawl ROS stacks.
         :param sources_loader: (optional) Override SourcesLoader used
             for managing sources.list data sources.
+        :param dependency_types: (optional) List of dependency types.
+            Allowed: "build", "buildtool", "run", "test"
```
mateus-amarante, line 343:
```suggestion
            Allowed: "build", "buildtool", "run", "test", "exec", "doc"
```
It also supports exec and doc dependencies

src/rosdep2/main.py
```diff
@@ -123,16 +123,21 @@ class UsageError(Exception):
...
+_global_options = None
+
+def _get_default_RosdepLookup(options=None):
     """
     Helper routine for converting command-line options into
     appropriate RosdepLookup instance.
     """
+    global _global_options
+    if options is None:
+        options = _global_options
```
mateus-amarante, line 135: This is no longer needed
```suggestion
```

src/rosdep2/main.py
```diff
@@ -123,16 +123,21 @@ class UsageError(Exception):
 """
 
 
-def _get_default_RosdepLookup(options):
+_global_options = None
+
+def _get_default_RosdepLookup(options=None):
```
mateus-amarante, line 128: I think it is better to keep this
```suggestion
def _get_default_RosdepLookup(options):
```

src/rosdep2/main.py
```diff
@@ -123,16 +123,21 @@ class UsageError(Exception):
 """
 
 
-def _get_default_RosdepLookup(options):
+_global_options = None
```
mateus-amarante, line 126:
```suggestion
_global_options = None
```
This is not needed

mateus-amarante (commented):
There are some problems with the last merge. The following suggestions might fix most of the problems.

Makefile
nuclearsandwich: If changes to the test runners are proposed I'd prefer them in a separate PR for ease of review.

src/rosdep2/lookup.py
nuclearsandwich: Should these docs also link to the specification for dependency types either in catkin_pkg or the ROS REPs defining the package.xml format?

nuclearsandwich (commented):
Overall these changes are looking promising. Thank you very much for adding tests along with this new feature.

I generally agree with the review suggestions from @mateus-amarante, would you have the opportunity to respond to that feedback before another round of review?

Makefile
mateus-amarante: Originally, the tests use `nosetests` (Python 2 version). I didn't intend to propose a test runner change. I changed it locally to try some stuff and committed it unintentionally. This suggestion reverts this change.

src/rosdep2/lookup.py
mateus-amarante: Good point. The `doc_depend` seems to be introduced in format 2 ( [REP127](https://ros.org/reps/rep-0127.html), [schema files](https://github.com/ros-infrastructure/rep/blob/master/xsd/package_format1.xsd) ), so maybe it is better not to consider it.

mateus-amarante:
@nuclearsandwich, if you agree, I can create a new PR from my fork, so that I can finish the updates and you can do another review. I'm very interested in this feature, so I will be committed to fixing any problem pointed out.

nuclearsandwich:
> I can create a new PR from my fork, so that I can finish the updates and you can do another review. I'm very interested in this feature, so I will be committed to fixing any problem pointed out.

Please go ahead!

emersonknapp:
@nuclearsandwich i think we can close this PR, #789 is looking good (and could use maintainer review :))

nuclearsandwich:
Replaced by #789

PR #729, opened 2019-11-12
openSUSE package query  and install enhancements

gleichdick:
Enable pip installer, use rpm capabilities for package lookup to prevent package renaming issues.

For example, in python.yml the `python-numpy` dependency is resolved to the opensuse rpm package `python-numpy` which is absent. But the package `python2-numpy` provides it as a capability as shown below:
```
$ rpm -q --provides python2-numpy
python-numpy = 1.16.5
python2-numpy = 1.16.5-2.1
python2-numpy(x86-64) = 1.16.5-2.1
```

So instead of matching the exact package name at dependency resolving the capabilities should be queried (`--whatprovides` flag).

```
$ rpm -q --whatprovides python2-numpy
python2-numpy-1.16.5-2.1.x86_64
$ rpm -q --whatprovides python-numpy
python2-numpy-1.16.5-2.1.x86_64
```

Of course, if none of the installed package provides the given capability the return code is still non-zero:

```
$ rpm -q --whatprovides anyunknowncap; echo $?
no package provides anyunknowncap
1
```

### DIFF
--- a/src/rosdep2/platforms/opensuse.py
+++ b/src/rosdep2/platforms/opensuse.py
@@ -31,6 +31,7 @@ import subprocess
 
 from rospkg.os_detect import OS_OPENSUSE
 
+from .pip import PIP_INSTALLER
 from .source import SOURCE_INSTALLER
 from ..installers import PackageManagerInstaller
 
@@ -44,6 +45,7 @@ def register_installers(context):
 
 def register_platforms(context):
     context.add_os_installer_key(OS_OPENSUSE, SOURCE_INSTALLER)
+    context.add_os_installer_key(OS_OPENSUSE, PIP_INSTALLER)
     context.add_os_installer_key(OS_OPENSUSE, ZYPPER_INSTALLER)
     context.set_default_os_installer_key(OS_OPENSUSE, lambda self: ZYPPER_INSTALLER)
 
--- a/src/rosdep2/platforms/opensuse.py
+++ b/src/rosdep2/platforms/opensuse.py
@@ -53,7 +53,7 @@ def register_platforms(context):
 def rpm_detect(packages):
     installed = []
     for p in packages:
-        if not subprocess.call(['rpm', '-q', p]):
+        if not subprocess.call(['rpm', '-q', '--whatprovides', p]):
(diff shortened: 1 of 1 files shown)


src/rosdep2/platforms/opensuse.py
```diff
@@ -44,14 +45,15 @@ def register_installers(context):
...
+    context.add_os_installer_key(OS_OPENSUSE, PIP_INSTALLER)
     context.add_os_installer_key(OS_OPENSUSE, ZYPPER_INSTALLER)
     context.set_default_os_installer_key(OS_OPENSUSE, lambda self: ZYPPER_INSTALLER)
 
 
 def rpm_detect(packages):
     installed = []
     for p in packages:
-        if not subprocess.call(['rpm', '-q', p]):
+        if not subprocess.call(['rpm', '-q', '--whatprovides', p]):
```
nuclearsandwich, line 56: It looks like the rpm facilities for the redhat platform are a bit more sophisticated. Is it worth centralizing and sharing some of this logic or is there too much divergence between them? Also asking @cottsay as my local RedHat expert and probable contributor for the redhat RPM handling.

cottsay: This is a similar approach to the one taken in the RedHat installer:
https://github.com/ros-infrastructure/rosdep/blob/cfa21c89d440e3fcb6c41b1f4a5f4947a5649b41/src/rosdep2/platforms/redhat.py#L110-L118

Invoking the `rpm` executable repetitively on a large workspace was pretty slow, so a single command is invoked for all of the packages and the output is parsed to ensure availability since the return value isn't useful anymore.

This change, however, shouldn't make the openSUSE installer any less performant than it currently is, and uses an already proven approach.

cottsay: > Is it worth centralizing and sharing some of this logic...?

Maybe. The RedHat installer tries to use the RPM python module if it is available, so I think we'd need to do some testing to ensure it works right, but I expect that the RPM solution in `redhat.py` would work here as-is.

I don't think that work should block this PR though.

PR #734, opened 2019-12-13
Add Suite3 option with Ubuntu Focal

sloretz:
Mimicking ros-infrastructure/catkin_pkg#271


dirk-thomas:
A separate PR should add Travis coverage for Python 3.7 and 3.8.

sloretz:
> A separate PR should add Travis coverage for Python 3.7 and 3.8.

Opened #735. Why does it need to be a separate PR? What would have happened if I put both changes into the same PR?

dirk-thomas:
> Why does it need to be a separate PR? What would have happened if I put both changes into the same PR?

Just the preference to keep two unrelated changes in separate PRs - no technical reason.

PR #737, opened 2020-01-04
Fix conditional dependencies when one package uses manifest.xml

### LINKED ISSUE #736: rosdep install python3 dependencies when a package has manifest.xml and package.xml
## Issue description

It appears that wrongly rosdep installs python3 version of packages, removing needed python2 packages in the process.

## How to reproduce

Trying to build `ros_comm` from source for Melodic:

To make the example clearer we'll install `python-catkin-tools` and build the workspace with that Python2 package

```bash
docker run -it --rm ubuntu:bionic
apt -qq update && apt -qq install -y lsb-release gnupg2 dirmngr
sh -c 'echo "deb http://packages.ros.org/ros/ubuntu $(lsb_release -sc) main" > /etc/apt/sources.list.d/ros-latest.list'
apt-key adv --keyserver 'hkp://keyserver.ubuntu.com:80' --recv-key C1CF6E31E6BADE8868B172B4F42ED6FBAB17C654
apt update && apt-get install python-rosdep python-rosinstall-generator python-wstool python-rosinstall build-essential python-catkin-tools
rosdep init && rosdep update
mkdir ~/ros_catkin_ws && cd ~/ros_catkin_ws
rosinstall_generator ros_comm --rosdistro melodic --deps --tar > melodic-ros_comm.rosinstall
wstool init -j8 src melodic-ros_comm.rosinstall
rosdep install --from-paths src --ignore-src --rosdistro melodic -y
catkin config --install
```

This fails with 
```
root@083919565a76:~/ros_catkin_ws# catkin config --install
bash: catkin: command not found
```

catkin tools has been uninstall by rosdep
```
executing command [apt-get install -y python3-catkin-pkg]
Reading package lists... Done
Building dependency tree       
Reading state information... Done
The following packages were automatically installed and are no longer required:
  javascript-common libjs-jquery libjs-sphinxdoc libjs-underscore python-catkin-pkg-modules python-concurrent.futures python-osrf-pycommon python-pyparsing python-rosdistro python-rosdistro-modules
  python-rospkg python-rospkg-modules python-setuptools python-trollius sudo
Use 'apt autoremove' to remove them.
The following additional packages will be installed:
  python3-catkin-pkg-modules python3-dateutil python3-docutils python3-olefile python3-pil python3-pkg-resources python3-pygments python3-pyparsing python3-roman python3-six
Suggested packages:
  fonts-linuxlibertine | ttf-linux-libertine texlive-lang-french texlive-latex-base texlive-latex-recommended python-pil-doc python3-pil-dbg python3-setuptools ttf-bitstream-vera python-pyparsing-doc
The following packages will be REMOVED:
  python-catkin-pkg python-catkin-tools python-rosdep python-rosdep-modules python-rosinstall python-rosinstall-generator
The following NEW packages will be installed:
  python3-catkin-pkg python3-catkin-pkg-modules python3-dateutil python3-docutils python3-olefile python3-pil python3-pkg-resources python3-pygments python3-pyparsing python3-roman python3-six
0 upgraded, 11 newly installed, 6 to remove and 0 not upgraded.
```

### Additional info
This issue does not happen if the `manifest.xml` from roslisp is not present

Possible reason: roslisp's `manifest.xml` depends on `rospack`, rosdep installs all dependencies of `rospack`, regardless of their ROS python version

sloretz:
Fixes #736

When rosdep finds a `manifest.xml` it uses `rospkg` to recursively find rosdep keys. `rospkg` doesn't evaluate conditional dependencies, so all rosdep keys get returned even if packages use a `package.xml`. This PR works around that by using `rospkg` for the first level of rosdeps, and then recursively using the logic to choose between `catkin_pkg` and `rospkg`.

### DIFF
--- a/src/rosdep2/rospkg_loader.py
+++ b/src/rosdep2/rospkg_loader.py
@@ -143,7 +143,13 @@ class RosPkgLoader(RosdepLoader):
             deps = pkg.build_depends + pkg.buildtool_depends + pkg.run_depends + pkg.test_depends
             return [d.name for d in deps if d.evaluated_condition]
         elif resource_name in self.get_loadable_resources():
-            return self._rospack.get_rosdeps(resource_name, implicit=implicit)
+            rosdeps = set(self._rospack.get_rosdeps(resource_name, implicit=False))
+            if implicit:
+                # This resource is a manifest.xml, but it might depend on things with a package.xml
+                # Make sure they get a chance to evaluate conditions
+                for dep in self._rospack.get_depends(resource_name):
+                    rosdeps = rosdeps.union(set(self.get_rosdeps(dep, implicit=True)))
+            return rosdeps
         elif resource_name in self._rosstack.list():
             # stacks currently do not have rosdeps of their own, implicit or otherwise
             return []
--- a/src/rosdep2/rospkg_loader.py
+++ b/src/rosdep2/rospkg_loader.py
@@ -149,7 +149,7 @@ class RosPkgLoader(RosdepLoader):
                 # Make sure they get a chance to evaluate conditions
                 for dep in self._rospack.get_depends(resource_name):
                     rosdeps = rosdeps.union(set(self.get_rosdeps(dep, implicit=True)))
-            return rosdeps
+            return list(rosdeps)
         elif resource_name in self._rosstack.list():
             # stacks currently do not have rosdeps of their own, implicit or otherwise
             return []


sloretz:
I didn't look too closely at the `rospkg` code. Maybe an alternative fix is to make `rospkg` evaluate conditions [here](https://github.com/ros-infrastructure/rospkg/blob/77d00ebbf8abeb252c813c2d4752db68675fa2a9/src/rospkg/manifest.py#L395-L396)?

sloretz:
@wjwwood FYI

wjwwood:
Also, can you have a look at the single test failure to see if it is related? For python 2.7 at least it appears to be passing on master but not this pr.

nuclearsandwich:
> I didn't look too closely at the `rospkg` code. Maybe an alternative fix is to make `rospkg` evaluate conditions [here](https://github.com/ros-infrastructure/rospkg/blob/77d00ebbf8abeb252c813c2d4752db68675fa2a9/src/rospkg/manifest.py#L395-L396)?

My gut feeling is that we should push this down as deep as we can so that we can think about conditions in as few places as possible. But to do so we'd need to pass the conditions through to the rospkg handler to forward to catkin_pkg?

nuclearsandwich:
@sloretz I pushed 0e28606 to address the test failure due to the export of an empty set (and sets in general) rather than lists. I don't *think* that it breaks anything within the recursive logic because we convert the interal call output to a set in the union() but I'd be grateful for a second set of eyes.

sloretz:
@nuclearsandwich 0e28606 LGTM

PR #749, opened 2020-04-03
Fix console script

nuclearsandwich:
Rebase of #656 onto master for CI. Maintainer pushes weren't turned on for that fork.


nuclearsandwich:
I'd rather review and merge #656

PR #750, opened 2020-04-03
Depend on modules packages only to allow co-installability.

nuclearsandwich:
When the rosdep modules package was split in [#731] the module
dependencies weren't updated to only depend on modules packages
themselves which prevents rosdep modules for python2 and python3 from
actually being co-installable.

I haven't yet audited to make sure there's no cli usage of these tools
in the modules but it's Friday afternoon and I wanted to get this inked
before I walked away.

When I return to it I'll check for that before proceeding for reviews.

### DIFF
--- a/stdeb.cfg
+++ b/stdeb.cfg
@@ -16,8 +16,8 @@ X-Python3-Version: >= 3.4
 Setup-Env-Vars: SKIP_PYTHON_MODULES=1
 
 [rosdep_modules]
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
CI won't catch anything here since the modules aren't split for a pip/setuptools installation.

A quick egrep for `subprocess\..*(rospkg|rosdistro|catkin)` didn't turn up any hits but that's not a very thorough check.

nuclearsandwich:
Alright a _slightly_ more thorough perusal of the sources leaves me somewhat confident that we're not calling any of the binaries in the rosdep modules sources.

PR #751, opened 2020-04-03
Use GitHub Actions for CI

nuclearsandwich:
This is a pretty straight-ish port of our Travis CI coverage to GitHub actions. Unlike Travis's matrix strategy which allows the specification of exact combinations with GitHub's you can only exclude or modify (via include) combinations from the set cartesian product. The setup-python action from upstream doesn't support Python 3.4 so I wasn't able to preserve support for it.

You can see this GitHub action running on my fork here: https://github.com/nuclearsandwich/rosdep/actions?query=workflow%3Arosdep-ci

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,8 +1,6 @@
 name: rosdep
 
-on:
-- push: [action-jack, master]
-- pull_request: [action-jack, master]
+on: [push, pull_request]
 
 jobs:
     build:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,8 +1,8 @@
 name: rosdep
 
 on:
-  push: [action-jack, master]
-  pull_request: [action-jack, master]
+- push: [action-jack, master]
+- pull_request: [action-jack, master]
 
 jobs:
     build:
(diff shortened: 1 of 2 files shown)


dirk-thomas:
While the package still targets platforms with Python 3.4 (like `Jessie`) I don't think it should be dropped from CI.

nuclearsandwich:
> While the package still targets platforms with Python 3.4 (like `Jessie`) I don't think it should be dropped from CI.

I think dropping Jessie for future releases is worth considering. Even the Debian LTS project support for Jessie is up in June.

dirk-thomas:
> I think dropping Jessie for future releases is worth considering. Even the Debian LTS project support for Jessie is up in June.

When doing so (likely after June) please make sure to do so consistently across all Python packages.

nuclearsandwich:
I've just updated this PR to keep the Travis build for Python 3.4. This has the trade-off of needing to make sure CI behavior between the two is kept in sync but it means that the majority of CI builds will benefit from improved reliability given Travis's recent and growing troubles. It also provides a good foothold for the work started in #661 to get reworked into GitHub Actions.

nuclearsandwich:
I noodled a bunch last night trying to add Python 3.9 and Ubuntu 20.04 to the test matrix and ended up punting on 20.04 for now since my local dev environment (Arch Linux) is passing on tests which are failing when running on Ubuntu. I'll get into a 20.04 environment eventually to investigate that but I'd rather forge ahead with other rosdep maintenance.

PR #753, opened 2020-04-06
also install buildtool_export_depends

### LINKED ISSUE #752: rosdep doesnt install buildtool_export_depend dependencies
This is a similar issue as https://github.com/ros-infrastructure/rosinstall_generator/pull/54 but for rosdep.

rosdep currently doesnt install the `buildtool_export` dependencies. This has been causing some build failures because of missing dependencies: e.g. https://github.com/ros2/sros2/runs/562579313?check_suite_focus=true

mikaelarguedas:
not adding 'exec_depends' or 'build_export_depends' are both are included inside 'run_depends'

Fixes #752 

Signed-off-by: Mikael Arguedas <mikael.arguedas@gmail.com>

### DIFF
--- a/src/rosdep2/rospkg_loader.py
+++ b/src/rosdep2/rospkg_loader.py
@@ -140,7 +140,7 @@ class RosPkgLoader(RosdepLoader):
         if resource_name in self.get_catkin_paths():
             pkg = catkin_pkg.package.parse_package(self.get_catkin_paths()[resource_name])
             pkg.evaluate_conditions(os.environ)
-            deps = pkg.build_depends + pkg.buildtool_depends + pkg.run_depends + pkg.test_depends
+            deps = pkg.build_depends + pkg.buildtool_depends + pkg.run_depends + pkg.test_depends + pkg.buildtool_export_depends
             return [d.name for d in deps if d.evaluated_condition]
         elif resource_name in self.get_loadable_resources():
             rosdeps = set(self._rospack.get_rosdeps(resource_name, implicit=False))


mikaelarguedas:
@nuclearsandwich can this be a candidate for the release underway? This is a rather serious bug currently preventing building ROS 2 workspaces on systems that don't already have python libraries installed. I opened this PR seeing that no release had been made for a while with the hope that this could be integrated before cutting a new release.

nuclearsandwich:
> @nuclearsandwich can this be a candidate for the release underway?

This release is already *stacked*. Since most of the effort is PR review and QA I'd just as well prefer to cut a follow-up 0.19.1 once the PR has been merged.

nuclearsandwich:
The code change is easy enough to review and there's no problem there.

There was quite some discussion about [rosdep handling doc_depends](https://github.com/ros-infrastructure/rosdep/pull/646) and which dependencies should be installed by default. I think the case for buildtool_export_depends being a default is in keeping with our other defaults, especially as the parallel fix in rosinstall_generator has been merged for some time. But I also want to call attention to #727 which I'd like to get wide feedback on, especially from folks like you @mikaelarguedas. Any change to the defaults will upset someone's apple cart and I think it would be good to have something in the pipeline so that different users can pick from available dependency types.

mikaelarguedas:
For sure I'm strongly in favor of #727 since it was brought to my attention at https://discourse.ros.org/t/generating-dev-and-runtime-artefacts-from-ros-packages/12448 :+1: I didn't get a chance to test it though.

> This release is already stacked. Since most of the effort is PR review and QA I'd just as well prefer to cut a follow-up 0.19.1 once the PR has been merged.

Do you think this has a chance to happen today ? or is it better to roll-back and release changes that exposed this problem in the meantime ?

dirk-thomas:
Related to ros-infrastructure/rospkg#191.

mikaelarguedas:
> I'm not sure that a decision will be made but I'm hoping we'll at least determine what info is needed to make an informed decision.

Sounds good

> The last comment in the other thread was asking the contributor to do a deeper assessment of the impact of the proposed change. I'm not sure that a decision will be made but I'm hoping we'll at least determine what info is needed to make an informed decision.

From my perspective, it makes sense to treat `buildtool_depends` and `buildtool_export_depends` the same way as they're both for the build platform and necessary for the build tool to successfully build a workspace.
Regarding if they should both be installed by default, I couldn't say. If the default/main use of rosdep is to get all dependences to build workspaces from source then most likely yes. If it's meant to be used on a target system to install runtime dependencies, then most likely no. But in that case a way to request all the deps needed for building workspaces should be provided.

nuclearsandwich:
> From my perspective, it makes sense to treat `buildtool_depends` and `buildtool_export_depends` the same way as they're both for the build platform and necessary for the build tool to successfully build a workspace

This was the convincing argument for me. In today's discussion we agreed that rosdep should install buildtool_export_depends by default along with buildtool_depends.

Using rosdep with distinct host and target platforms is not a use case that rosdep currently supports but a PR like #727 could enable it in the future.

mikaelarguedas:
@nuclearsandwich Do you know when we can expect the follow-up release of rosdep including this fix ? :pray:

ros-discourse:

https://discourse.ros.org/t/announcement-trimmed-ros-gazebo-docker-images-released/14342/1

mikaelarguedas:
@nuclearsandwich friendly ping, is there a timeframe available for getting this fix released ?

nuclearsandwich:
Short answer no. I won't be able to get to a rosdep release this week or possibly next. But I'll try for the week after.

mikaelarguedas:
@nuclearsandwich friendly ping. This seem to still cause issues for people attenmpting to compile ROS2 from source using rosdep to install their workspace dependencies.

Is there a chance of getting this released soon ?

PR #771, opened 2020-07-23
If sources list file has default content, exit gracefully

rotu:
Fixes https://github.com/ros-tooling/setup-ros/issues/174

### DIFF
--- a/src/rosdep2/main.py
+++ b/src/rosdep2/main.py
@@ -595,7 +595,7 @@ def command_init(options):
         path = get_default_sources_list_file()
         try:
             with open(path, 'r') as f:
-                if f.readlines() == data.splitlines():
+                if f.read().splitlines() == data.splitlines():
                     print('Default sources list file okay. Nothing to do.')
                     return 0
                 else:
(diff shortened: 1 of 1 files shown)


emersonknapp (approved):
Nice!

rotu:
What’s my next step to get this merged in?

rotu:
What’s my next step to get this merged in?

emersonknapp:
Maybe we ping @nuclearsandwich @tfoote for review?

nuclearsandwich:
My initial reaction was that invoking rosdep init in a script like

```
test -f /etc/ros/rosdep/sources.list.d/20-default.list || rosdep init
```

would be a reasonable way to get this feature without a code change but that script assumes the default path would never change and doesn't warn if the contents update so this has obvious advantages.

@rotu would you be interested in adding tests for the two reinitialization cases? (same content and changed content)

nuclearsandwich (changes requested):
I think all this needs is tests and then we can merge it.

rotu:
@nuclearsandwich Feel free to take over this PR

nuclearsandwich:
> @nuclearsandwich Feel free to take over this PR

Thanks, I'm not sure I'll get to it but if anyone wants to see this change merged open a PR with these changes and some tests and I'll review it.

PR #776, opened 2020-08-21
Fix handling of installer version strings

christophebedard:
The `--all-versions` option currently fails:

```
0.19.0

ERROR: Rosdep experienced an error: a bytes-like object is required, not 'str'

rosdep version: 0.19.0

  File "/home/chris/rosdep/src/rosdep2/main.py", line 144, in rosdep_main
  File "/home/chris/rosdep/src/rosdep2/main.py", line 389, in _rosdep_main
    installer_version_strings = installer.get_version_strings()
  File "/home/chris/rosdep/src/rosdep2/platforms/debian.py", line 266, in get_version_strings
    version = output.splitlines()[0].split(' ')[1]
TypeError: a bytes-like object is required, not 'str'
```

I assume this (change to `bytes`) wasn't caught during the move to Python 3.

I also checked the other implementations of `get_version_strings()` and added calls to `decode()` where necessary. Otherwise it prints out `b'1.2.3'` instead of `1.2.3`. I haven't tested nix or opkg, but it works fine for gem and apt-get.

### DIFF
--- a/src/rosdep2/platforms/debian.py
+++ b/src/rosdep2/platforms/debian.py
@@ -263,7 +263,7 @@ class AptInstaller(PackageManagerInstaller):
 
     def get_version_strings(self):
         output = subprocess.check_output(['apt-get', '--version'])
-        version = output.splitlines()[0].split(' ')[1]
+        version = output.splitlines()[0].split(b' ')[1].decode()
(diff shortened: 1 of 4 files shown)


nuclearsandwich (approved):
Thanks very much!

PR #785, opened 2020-11-13
Restrict on-push triggers to default branch.

nuclearsandwich:
Follow-up to #751 to prevent duplicate runs when pushing branches.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,6 +1,10 @@
 name: rosdep-ci
 
-on: [push, pull_request]
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: ['*']
 
 jobs:
     build:


.github/workflows/ci.yaml
```diff
@@ -1,6 +1,10 @@
 name: rosdep-ci
 
-on: [push, pull_request]
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: ['*']
```
cottsay, line 7: I think this line can be omitted. Just the `pull_request:` has worked for me in the past. Not a big deal - if you want to keep it for clarity, that's fine with me.

cottsay (approved):

.github/workflows/ci.yaml
nuclearsandwich: Thanks for the comment. And yeah empty yaml keys always make me a bit nervous so I like that I can put something in this one.

nuclearsandwich:
I'm assuming the travis job will get there eventually.

Issue #782, opened 2020-10-16
ERROR: Rosdep experienced an error: [Errno 13] Permission denied:


iayushguptaaa:
rosdep version: 0.19.0

Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/rosdep2/main.py", line 144, in rosdep_main
    exit_code = _rosdep_main(args)
  File "/usr/lib/python2.7/dist-packages/rosdep2/main.py", line 430, in _rosdep_main
    return _package_args_handler(command, parser, options, args)
  File "/usr/lib/python2.7/dist-packages/rosdep2/main.py", line 526, in _package_args_handler
    lookup = _get_default_RosdepLookup(options)
  File "/usr/lib/python2.7/dist-packages/rosdep2/main.py", line 134, in _get_default_RosdepLookup
    verbose=options.verbose)
  File "/usr/lib/python2.7/dist-packages/rosdep2/sources_list.py", line 594, in create_default
    sources = load_cached_sources_list(sources_cache_dir=sources_cache_dir, verbose=verbose)
  File "/usr/lib/python2.7/dist-packages/rosdep2/sources_list.py", line 552, in load_cached_sources_list
    with open(cache_index, 'r') as f:
IOError: [Errno 13] Permission denied: '/home/bingo/.ros/rosdep/sources.cache/index'

nuclearsandwich:
Does the error persist if you run `sudo rosdep fix-permissions`? If `rosdep update` is run with `sudo` after running `sudo rosdep init` it the created cache files in your home directory will be owned by root.

nuclearsandwich:
#787 improves the error message here to match other situations where we detect permissions issues.

PR #787, opened 2020-11-14
Show rosdep-fix-permissions advice when cache read fails.

nuclearsandwich:
By re-raising this IOError as a CachePermissionsError the advice to run
rosdep fix-permissions will be displayed instead of the general
stacktrace.

If rosdep update is run as root and then a command like rosdep resolve
is used to read from the cache rather than write to it this stacktrace
occurs as reported in #782.

### Before

```
❯ rosdep resolve python

ERROR: Rosdep experienced an error: [Errno 13] Permission denied: '/home/steven/.ros/rosdep/sources.cache/index'

rosdep version: 0.20.0

  File "/home/steven/osrf/rosdep/src/rosdep2/main.py", line 144, in rosdep_main
  File "/home/steven/osrf/rosdep/src/rosdep2/main.py", line 426, in _rosdep_main
    return _rosdep_args_handler(command, parser, options, args)
  File "/home/steven/osrf/rosdep/src/rosdep2/main.py", line 448, in _rosdep_args_handler
    return command_handlers[command](args, options)
  File "/home/steven/osrf/rosdep/src/rosdep2/main.py", line 853, in command_resolve
    lookup = _get_default_RosdepLookup(options)
  File "/home/steven/osrf/rosdep/src/rosdep2/main.py", line 132, in _get_default_RosdepLookup
    sources_loader = SourcesListLoader.create_default(sources_cache_dir=options.sources_cache_dir,
  File "/home/steven/osrf/rosdep/src/rosdep2/sources_list.py", line 603, in create_default
    sources = load_cached_sources_list(sources_cache_dir=sources_cache_dir, verbose=verbose)
  File "/home/steven/osrf/rosdep/src/rosdep2/sources_list.py", line 561, in load_cached_sources_list
    with open(cache_index, 'r') as f:
PermissionError: [Errno 13] Permission denied: '/home/steven/.ros/rosdep/sources.cache/index'
```

### After

```
❯ rosdep resolve python
Failed to write cache file: [Errno 13] Permission denied: '/home/steven/.ros/rosdep/sources.cache/index'
Try running 'sudo rosdep fix-permissions'
```

### DIFF
--- a/src/rosdep2/sources_list.py
+++ b/src/rosdep2/sources_list.py
@@ -558,8 +558,14 @@ def load_cached_sources_list(sources_cache_dir=None, verbose=False):
         if verbose:
             print('no cache index present, not loading cached sources', file=sys.stderr)
         return []
-    with open(cache_index, 'r') as f:
-        cache_data = f.read()
+    try:
+        with open(cache_index, 'r') as f:
+            cache_data = f.read()
+    except IOError as e:
+        if e.strerror == 'Permission denied':
+            raise CachePermissionError('Failed to write cache file: ' + str(e))
+        else:
+            raise
     # the loader does all the work
     model = cache_data_source_loader(sources_cache_dir, verbose=verbose)
     return parse_sources_data(cache_data, origin=cache_index, model=model)
--- a/src/rosdep2/sources_list.py
+++ b/src/rosdep2/sources_list.py
@@ -563,7 +563,7 @@ def load_cached_sources_list(sources_cache_dir=None, verbose=False):
             cache_data = f.read()
     except IOError as e:
         if e.strerror == 'Permission denied':
-            raise CachePermissionError('Failed to write cache file: ' + str(e)) from e
+            raise CachePermissionError('Failed to write cache file: ' + str(e))
         else:
             raise
     # the loader does all the work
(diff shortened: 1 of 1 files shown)


src/rosdep2/sources_list.py
```diff
@@ -558,8 +558,14 @@ def load_cached_sources_list(sources_cache_dir=None, verbose=False):
...
             print('no cache index present, not loading cached sources', file=sys.stderr)
         return []
-    with open(cache_index, 'r') as f:
-        cache_data = f.read()
+    try:
+        with open(cache_index, 'r') as f:
+            cache_data = f.read()
+    except IOError as e:
+        if e.strerror == 'Permission denied':
+            raise CachePermissionError('Failed to write cache file: ' + str(e))
```
cottsay, line 566: It would be a good idea to preserve the original context here.
```suggestion
            raise CachePermissionError('Failed to write cache file: ' + str(e)) from e
```

cottsay (approved):
One minor suggestion, but this looks good to me.

src/rosdep2/sources_list.py
nuclearsandwich: This appears to cause syntax errors in python2. rosdep doesn't currently depend on six or any other such wrappers that would bridge this gap for us.

It seems like `exec("raise...")` could do what we want for only python3. I'm not afraid of it but it'll be the first time that exec/eval are used in rosdep outside of some tests.

cottsay: I completely forgot that this wasn't valid in Python 2. Just forget about it - it was just a Nice To Have.

PR #789, opened 2020-11-16
Selection of dependency types with cmd line option [#727 continuation]

mateus-amarante:
This is a refactor of #727 update (started by @nalt) which includes adding `--dependency-types` / `-t` cmd line option to select the dependency types for commands like `install`, `key` and `check`. It supports specification of "build", "buildtool", "build_export", "run", "test", "exec" and "doc" types. This PR also adds related tests.


mateus-amarante:
@nuclearsandwich friendly ping. This is the realization of https://github.com/ros-infrastructure/rosdep/pull/727#issuecomment-728140397

emersonknapp:
Should #727 be closed?

src/rosdep2/main.py
```diff
@@ -368,6 +368,11 @@ def _rosdep_main(args):
                       help="Affects the 'update' verb. "
                            'If specified end-of-life distros are being '
                            'fetched too.')
+    parser.add_option('-t', '--dependency-types', dest='dependency_types',
+                      type="choice", choices=("build", "buildtool", "build_export", "exec", "run", "test", "doc"),
```
emersonknapp, line 372: should this be `nargs='+'` - which should let you omit the `default=[]` because it will be an empty list when not specified. I don't think that `type="choice"` is correct - shouldn't it be `type=str` (which I think is the default)?

Additionally (just thinking out loud) - is there any place that we could query these choices from rather than hardcoding them here? May be better to have a single source of truth for the types of dependencies that are allowed, and if it ever changes this code wouldn't rot.

src/rosdep2/rospkg_loader.py
```diff
@@ -78,6 +78,20 @@ def __init__(self, rospack=None, rosstack=None, underlay_key=None):
...
         self._catkin_packages_cache = None
 
+        # Dependency types to include
+        def check_dep(dep_type):
+            if dependency_types:
+                return dep_type in dependency_types
+            else:
+                return dep_type in ['buildtool', 'build', 'build_export', 'exec', 'test']
+
+        self.include_build_depends = check_dep('build')
```
emersonknapp, line 88: this could probably be

```
# these should probably be constants defined in one place and used across this PR
all_dep_types = {'buildtool', 'build', 'build_export', 'exec', 'test', 'doc'}
default_dep_types = all_dep_types - {'doc'}

self.include_dep_types = all_dep_types.intersection(set(dependency_types)) if dependency_types else default_dep_types
```

then below in `get_rosdeps` we could use a generic iteration as well

```
deps = sum((getattr(pkg, f'{d}_depends') for d in self.include_dep_types), [])
```

In fact, after writing that, one way to get `all_dep_types` as above could be to introspect the `pkg` type object with something like

```
all_dep_types = [x for x in dir(pkg) if x.endswith('_depends')]
```

mateus-amarante: I was unhappy with this floating function... Very nice suggestions! I'll consider applying them.

src/rosdep2/main.py
mateus-amarante: You are right. It is better to have these names defined in a single place. Do you have a suggestion on where they should be?

emersonknapp: The source of truth seems to be `catking_pkg.package.Package` (https://github.com/ros-infrastructure/catkin_pkg/blob/master/src/catkin_pkg/package.py#L53) - which has the attribute `__slots__`, that might be the best place to parse these values from, though it may not be best practice to go grabbing dunder variables. I'm not super experienced with the source of this catkin infra, input from a maintainer would be helpful on the best approach.

src/rosdep2/rospkg_loader.py
mateus-amarante: Implemented this suggestion in fa3fd33. I still didn't find a good format for `all_dep_types`. For now, it is a function in rospkg_load.py.

mateus-amarante:
> Should #727 be closed?

Yes. This is a replacement.

src/rosdep2/main.py
```diff
@@ -368,6 +368,11 @@ def _rosdep_main(args):
                       help="Affects the 'update' verb. "
                            'If specified end-of-life distros are being '
                            'fetched too.')
+    parser.add_option('-t', '--dependency-types', dest='dependency_types',
+                      type="choice", choices=list(all_dep_types()),
+                      default=[], action='append',
+                      help='Dependency types to install, can be given multiple times. '
+                           'Choose from {}. Default: all except doc.'.format(all_dep_types()))
```
emersonknapp, line 375: This doesn't work as written, you should probably use the following options:

```suggestion
    parser.add_option('-t', '--dependency-types',
                      type=str, choices=list(all_dep_types()),
                      nargs='+',
                      help='Space-separated list of dependency types to install. Default: all except "doc"'
```

* nitpick: `dest` automatically replaces hyphens with underscores, this is a no-op
* `type` should name the type of argument being parsed - `"choice"` is not a type, `str` is: this raises an exception right now because `'choice' is not callable`
*  `nargs='+'` allows one or more arguments, defaulting to an empty list (meaning you don't need to specify `default` or `action`) - without this, if you pass in multiple choices, then the second one will raise `unrecognized argument: buildtool`. If you use `nargs='+', action='append'` in combination, then you'll actually get a nested list `dependency_types=[['build', 'exec']]` rather than the desired `dependency_types=['build', 'exec']`

mateus-amarante: Note the parser is from [optparse](https://docs.python.org/3/library/optparse.html), so `type` is a name from an extendable list of [options](https://docs.python.org/3/library/optparse.html#optparse-standard-option-types) and `nargs` is a fixed [number](https://docs.python.org/3/library/optparse.html#optparse.Option.nargs).

Currently, the user needs to repeat the arg keyword like `-t build -t exec -t test`. However, I agree it would be better to provide the list directly. To do this, we can do like [`--skip-keys`](https://github.com/ros-infrastructure/rosdep/blob/2977726b32f57205ac0985ce9eb8f44e5dbbe8bf/src/rosdep2/main.py#L404-L405) which expect a string with comma-separated names like `--skip-keys "pkg1 pkg2 pkg3"` or, even better, we could create a custom action like this [example](https://docs.python.org/3/library/optparse.html#adding-new-actions) that allows the user to specify comma-separated list  like `--dependency-types build,exec,test`

emersonknapp: Ah - that's on me for assuming argparse - sorry about that. I just didn't even think of it since it's been deprecated for a while. I  think space-separated is preferable - comma-separated is not a very common commandline paradigm.

emersonknapp:
ping @nuclearsandwich i'd love to have this feature (and i think the implementation is good now - and https://github.com/ros-infrastructure/rosdep/pull/727 can be closed in favor of this PR

test/test_rosdep_rospkg_loader.py
```diff
@@ -179,3 +179,8 @@ def test_RosPkgLoader_get_loadable():
     keys = loader.get_loadable_views()
     for s in ['ros', 'empty', 'invalid', 'stack1']:
         assert s in keys
+
+
+def test_rospkg_loader_all_dep_types():
+    from rosdep2.rospkg_loader import all_dep_types
+    assert all_dep_types() == {'build', 'buildtool', 'build_export', 'buildtool_export', 'exec', 'test', 'doc'}, all_dep_types()
```
130s, line 186: This is a test file so fixed values may make more sense, but I still want to raise this thoughts: Should the types be taken from the upstream `catkin_pkg.package.Package` like done [in the source](https://github.com/ros-infrastructure/rosdep/pull/789/files#diff-68811be333f5db97a55f12cb74a4a059af80d63e4d599c7a51ad104d647d6c93R62)?

test/tree/catkin/multi_dep_type_catkin_package/package.xml
```diff
@@ -0,0 +1,19 @@
+<package format="2">
```
130s, line 1:
```suggestion
<package format="3">
```
Maybe nit, and open to counter suggestion; my personal preference is new development to  happen in format 3.

130s (commented):
I came from [answers.ros.org#q370946](https://answers.ros.org/question/370946/avoid-test_depend-in-rosdep-install/?comment=370952#post-id-370952). Great feature!

test/test_rosdep_rospkg_loader.py
mateus-amarante: I think this strategy isn't adequate because it would use the same principle the function being tested uses as if I was doing `asssert all_dep_types() == all_dep_types()`, which would be always true. Imagine if a new dependency type is added to `Package` and we don't want to consider it just yet. The tests would pass without our intention.

On the other hand, you might argue that this list could be constant in the source code and the test could check the list against all the dep types in `Package`. I think this is valid too and maybe it is in better agreement with my [dissatisfaction](https://github.com/ros-infrastructure/rosdep/pull/789#discussion_r540844183) with this function. I think the maintainers could help us with this decision.

test/tree/catkin/multi_dep_type_catkin_package/package.xml
mateus-amarante: All the dependency types considered in the new feature were introduced in format 2. It is the minimum format required for this feature. Format 3 is compatible with it so it should work too (we could change it without breaking the tests). However, it is important to highlight that I'm ignoring the new dependency type `group_depend`, introduced in format 3. I was not sure how I would handle it, so I excluded it from this update.

If the current behavior is considered adequate for format 3 given the `group_depend` ignoring, I'm ok with changing it.

test/test_rosdep_rospkg_loader.py
130s: > On the other hand, you might argue that this list could be constant in the source code and the test could check the list against all the dep types in `Package`.

+1 for that (I'm not a maintainer).

gavanderhoorn:
@nuclearsandwich: what would it take to get this in? More reviews? More testing? is there anything else that needs attention here?

emersonknapp:
@wjwwood @tfoote are there alternative maintainers we should ping for review on this (such as you)? it seems that steven! is otherwise occupied

src/rosdep2/lookup.py
```diff
@@ -328,7 +328,7 @@ def get_resources_that_need(self, rosdep_name):
     @staticmethod
     def create_from_rospkg(rospack=None, rosstack=None,
                            sources_loader=None,
-                           verbose=False):
+                           verbose=False, dependency_types=[]):
```
wjwwood, line 331: I wouldn't recommend this because of mutable default arguments:

https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

wjwwood: Even if you only read from it.

wjwwood (commented):
I think I need to wait for another maintainer to have a look, or maybe test it. But I had a glance at it, hopefully that helps them merge with confidence.

src/rosdep2/lookup.py
emersonknapp: Wacky gotcha! Nice call-out

cottsay (changes requested):
I like the feature that this PR implements, but I'm not sold on the "default types" enumeration introduced by fa3fd33 for a couple of reasons:
1. I don't consider the use of `__slots__` to be part of catkin_pkg's API, but rather an implementation optimization.
2. It could lead to an unexpected change in behavior if additional dependency types are introduced in catkin_pkg. It's worth a separate discussion.

It's not that I think it's the wrong thing to do, but it's really a separate change from what this PR is trying to implement. I'd like to see it moved to a separate PR so that we can move forward with getting this one merged.

mateus-amarante:
@cottsay Thanks for the review!

See if 235131d is ok and let me know if it requires any more changes.

yotabits:
@cottsay Would it be possible to have some feedback on this? It would be a really great addition to rosdep. Thank you.

yotabits:
@cottsay @dirk-thomas @wjwwood friendly reminder :)

jacobperron (approved):
I'm not familiar with the rosdep code, but the changes look sound and I've tried it out and it's working in my case :+1:

wjwwood:
Seems reasonable to me, but up to the current rosdep maintainers.

emersonknapp:
@wjwwood can you remind us who the current maintainers are? There is no package.xml for rosdep - I was under the impression that is was you, @tfoote and @nuclearsandwich . 

This one's been in progress for about 18 months (if you include the original PR that this extends) - who has the final say? It sounds like everybody agrees that it looks fine.

gavanderhoorn:
> Seems reasonable to me, but up to the current rosdep maintainers.

Who are the maintainers? @tfoote and Ken Conley are listed as authors, but there is no maintainer specified for `rosdep`.

Looks like @nuclearsandwich and @cottsay have released newer versions in the recent past?

As others have mentioned, this PR seems to be the last piece needed to be able to start working towards (nicer) multi-stage Docker builds/images. For that use-case alone it would be very nice to get this in.

What's keeping this from getting merged right now? @mateus-amarante seems to have made every requested change and there are multiple positive reviews.

quentingllmt:
Friendly message to tell we are also waiting for this feature in the multi-stage Docker build context.
When could you merge this PR ? Thanks !

nuclearsandwich (commented):
After reviewing this morning I have one comment and after talking it through offline with @cottsay I'm going to implement my own suggestion and have him do a review of the result. Then we'll be in mergeable shape.

There's a lingering question / concern that both of us have about whether catkin-pkg or rosdep should "own" the list of possible dependency types but as of now catkin-pkg doesn't have an explicit API for them and we agree that maintaining a list in rosdep is going to carry us forward for now.

I think a big indicator that reorganization will be required is if the list of dependencies starts getting used in other projects it will make sense to expose it directly in catkin-pkg.

src/rosdep2/rospkg_loader.py
```diff
@@ -57,9 +57,13 @@
 # no longer resource-dependent.
 
 
+def all_dep_types():
+    return {'build', 'buildtool', 'build_export', 'buildtool_export', 'exec', 'test', 'doc'}
```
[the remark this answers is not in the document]
nuclearsandwich: Potentially resolved by 0434ac8

cottsay (approved):

gavanderhoorn:
:tada: :+1: 

thanks @nalt and @mateus-amarante

ruffsl:
@gavanderhoorn , I mocked up a change to the multistage example in the ros docs for the docker hub library image in order to take advantage of this new feature in rosdep. However when simply building and installing the talker listener demo, the final image size savings are not yet compelling: https://github.com/ruffsl/docs-1/pull/1

Do you have a ROS example use case that could better showcase or demonstrate the space savings of only installing exec dependencies' in a final runner multistage image layer?

PR #792, opened 2021-01-05
List only formula in brew_detect for brew 3.0.

smnogar:
I think its only a cosmetic change for Mac. Fixes `Error: Calling brew list to only list formulae is disabled! Use brew list --formula instead.` warning.

### DIFF
--- a/src/rosdep2/platforms/osx.py
+++ b/src/rosdep2/platforms/osx.py
@@ -184,7 +184,7 @@ def brew_detect(resolved, exec_fn=None):
     """
     if exec_fn is None:
         exec_fn = read_stdout
-    std_out = exec_fn(['brew', 'list'])
+    std_out = exec_fn(['brew', 'list', '--formula'])
     installed_formulae = std_out.split()
 
     def is_installed(r):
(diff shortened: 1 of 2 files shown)


nuclearsandwich:
@smnogar can you elaborate on when and where you see this error? I updated the PR to fix the failing tests but I'm not sure what changed in homebrew which makes this required.

smnogar:
To be honest I'm having trouble remembering the exact circumstances I was getting this. Homebrew has become a lot more hostile to installing and/or running non-current packages. I found similar issues over in [nvm](https://github.com/rvm/rvm/issues/5007) that provides discussion of the same issue. 

I haven't come across this issue since this was submitted, so maybe this was fixed by simply updating homebrew. You can disregard this PR if no one else is running into this.

nuclearsandwich:
Looking at the commit history for the list command it seems like it started getting enhanced features in order to retire the `brew cask` subcommand and that the warning message reported here was removed for brew 3.0.

So `brew list` is will still print formulae and casks, although it only prints casks if there's one installed, which is why I wasn't seeing this behavior on my test system. So I think we do want to take this update otherwise we run the risk of parsing errors when using brew on systems with configured casks.

nuclearsandwich (approved):

nuclearsandwich (approved):

Issue #513, opened 2017-05-13
OSX Unable to resolve dependencies [pyqt]


budekatude:
I've been following the installation instructions for Indigo on Mac OSX in the ROS wiki. Everything works fine until I try to resolve the dependencies:

`Matty:ros_catkin_ws Matthias$ rosdep install --from-paths src --ignore-src --rosdistro indigo -y`

```
ERROR: the following packages/stacks could not have their rosdep keys resolved
to system dependencies:
python_qt_binding: No definition of [python-qt-bindings] for OS [osx]
qwt_dependency: No definition of [python-qt-bindings-qwt5] for OS [osx]
rviz: No definition of [libqt4-opengl-dev] for OS [osx]
gl_dependency: No definition of [python-qt-bindings-gl] for OS [osx]
webkit_dependency: No definition of [python-qt-bindings-webkit] for OS [osx]
qt_gui_cpp: No definition of [qt4-qmake] for OS [osx]
qt_gui: No definition of [python-qt-bindings] for OS [osx]
turtlesim: No definition of [libqt4] for OS [osx]
Matty:ros_catkin_ws Matthias$ cd /usr/local/Cellar
Matty:Cellar Matthias$ sudo ln -s pyqt5 pyqt
```

I tried manually installing pyqt from homebrew and this is the status I get:

```
pyqt: stable 5.8.2 (bottled)
Python bindings for v5 of Qt
https://www.riverbankcomputing.com/software/pyqt/download5
/usr/local/Cellar/pyqt/5.8.2 (856 files, 24.7MB) *
  Poured from bottle on 2017-05-13 at 15:08:46
From: https://github.com/Homebrew/homebrew-core/blob/master/Formula/pyqt.rb
==> Dependencies
Required: qt ✔, sip --with-python3 ✔
==> Requirements
Recommended: python3 ✔
Optional: python ✔
==> Options
--with-debug
	Build with debug symbols
--with-docs
	Install HTML documentation and python examples
--with-python
	Build with python support
--without-python3
	Build without python3 support
```
I've also tried relinking pyqt, without resolving the problem.

Any help is highly appreciated, sorry if I missed something... fairly new to this.
Thanks.

wjwwood:
Indigo is not going to work any longer because Homebrew removed all of the `qt4` related formulae:

https://github.com/ros/homebrew-deps/pull/31

wjwwood:
[...]

Issue #490, opened 2016-11-24
OSX, brew install failed to detect successful installation of [qt5]


excelulous:
Running this command:
```
MacBook:ros_catkin_ws user$ rosdep install --from-paths src --ignore-src --rosdistro kinetic -y`
executing command [brew install qt5]
Warning: qt5-5.7.0 already installed
ERROR: the following rosdeps failed to install
   homebrew: Failed to detect successful installation of [qt5]
```

I also get this stacktrace, but I understand it might be caused by another, unrelated issue:
```
rosdep version: 0.11.5

Traceback (most recent call last):

 File "/usr/local/lib/python2.7/site-packages/rosdep2/main.py", line 136, in rosdep_main
    exit_code = _rosdep_main(args)
  File "/usr/local/lib/python2.7/site-packages/rosdep2/main.py", line 353, in _rosdep_main
    return _package_args_handler(command, parser, options, args)
  File "/usr/local/lib/python2.7/site-packages/rosdep2/main.py", line 447, in _package_args_handler
    return command_handlers[command](lookup, packages, options)
  File "/usr/local/lib/python2.7/site-packages/rosdep2/main.py", line 634, in command_install
    print("uninstalled dependencies are: [%s]"%(', '.join([', '.join(pkg) for pkg in [v for k,v in uninstalled]])))
TypeError: sequence item 0: expected string, HomebrewResolution found
```

I installed pyqt5 separately via this command:
`brew install pyqt5 --with-python`

And Homebrew tells me pyqt5 is installed:
```
brew info pyqt5
pyqt5: stable 5.7 (bottled)
Python bindings for v5 of Qt
https://www.riverbankcomputing.com/software/pyqt/download5
/usr/local/Cellar/pyqt5/5.7 (901 files, 42.9M) *
  Built from source on 2016-11-24 at 16:03:51 with: --with-python
From: https://github.com/Homebrew/homebrew-core/blob/master/Formula/pyqt5.rb
==> Dependencies
Required: qt5 ✔, sip --with-python3 ✔
==> Requirements
Recommended: python3 ✔
Optional: python ✔
==> Options
--with-debug
	Build with debug symbols
--with-docs
	Install HTML documentation and python examples
--with-python
	Build with python support
--without-python3
	Build without python3 support
```

Any idea what's going on?

tfoote:
How do you "also get this stacktrace"? it looks to be a duplicate of #486.

Also can you give the brew info for `qt5` which is the one failing.

[...]

PR #799, opened 2021-01-30
homebrew: accept keg-only packages as installed

### LINKED ISSUE #764: I have a problem using rosdep install command
rosdep install --from-paths src --ignore-src --rosdistro kinetic -y --as-root pip:no --skip-keys="python-qt-bindings-qwt5 libqt5-core libqt5-gui libqt5-opengl libqt5-opengl-dev libqt5-widgets qt5-qmake qtbase5-dev python-imaging" --skip-keys google-mock --skip-keys python-wxtools

this is the command I use; however, I get this error and deprecation warning :

Deprecation: Even though I have python version 3.8

DEPRECATION: Python 2.7 reached the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 is no longer maintained. pip 21.0 will drop support for Python 2.7 in January 2021. More details about Python 2 support in pip, can be found at https://pip.pypa.io/en/latest/development/release-process/#python-2-support

That is what I get when I run the above command:
executing command [brew install pyqt5]
Warning: pyqt 5.14.2 is already installed and up-to-date
To reinstall 5.14.2, run `brew reinstall pyqt`
  homebrew: Failed to detect successful installation of [pyqt5]

I also tried to unlink pyqt5 and link it again but it doesn't work and I also tried to uninstall pyqt5 and install it again separately but also it doesn't work.

please help!

HesselM:
fixes #764 #513 #490 (and others)

As suggested by @NikolausDemmel in his [comment](https://github.com/ros-infrastructure/rosdep/issues/490#issuecomment-276224759) :

The issue is that qt5 is keg-only and rosdep marks it as "not installed", when it in fact, it is installed yet not symlinked as mentioned by brew during (re)installation:

```
==> Reinstalling qt 
==> Pouring qt-5.15.2.catalina.bottle.tar.gz
==> Caveats
We agreed to the Qt open source license for you.
If this is unacceptable you should uninstall.

qt is keg-only, which means it was not symlinked into /usr/local,
because Qt 5 has CMake issues when linked.

If you need to have qt first in your PATH, run:
  echo 'export PATH="/usr/local/opt/qt/bin:$PATH"' >> /Users/Electroozz/.bash_profile

For compilers to find qt you may need to set:
  export LDFLAGS="-L/usr/local/opt/qt/lib"
  export CPPFLAGS="-I/usr/local/opt/qt/include"

For pkg-config to find qt you may need to set:
  export PKG_CONFIG_PATH="/usr/local/opt/qt/lib/pkgconfig"

```

As a result, Rosdep fails:
```
$ rosdep install -i --from-path src --rosdistro foxy -y
executing command [brew install qt5]
Warning: qt 5.15.2 is already installed and up-to-date.
To reinstall 5.15.2, run:
  brew reinstall qt
  homebrew: Failed to detect successful installation of [qt5]
```

By checking if a package is keg-only when no linked keg is found, packages like Qt are accepted and rosdep finished clean.

### DIFF
--- a/src/rosdep2/platforms/osx.py
+++ b/src/rosdep2/platforms/osx.py
@@ -204,7 +204,11 @@ def brew_detect(resolved, exec_fn=None):
             pkg_info = pkg_info[0]
             linked_version = pkg_info['linked_keg']
             if not linked_version:
-                return False
+                keg_only = pkg_info['keg_only']
+                if keg_only:
+                    return True
+                else:
+                    return False
             for spec in pkg_info['installed']:
                 if spec['version'] == linked_version:
                     installed_options = spec['used_options']


nuclearsandwich:
@HesselM thanks for opening this PR and proposing changes to improve the situation with homebrew. My main concern with the change as proposed is keg-only dependencies will require additional setup to be usable and rosdep may mask that fact by reporting keg-only packages as installed.

I think we definitely need keg-only packages to be recognized as installed but does it make sense to print an info message when a dependency is found as keg-only?

wjwwood:
I think it's important to have some notice to the user that the formula is keg-only and that rosdep hasn't resolved that issue, but maybe the caveat output of Homebrew during the rosdep install is sufficient?

I don't feel strongly about it.

wjwwood (approved):
lgtm once the conceptual question is resolved.

scpeters:
> I think it's important to have some notice to the user that the formula is keg-only and that rosdep hasn't resolved that issue, but maybe the caveat output of Homebrew during the rosdep install is sufficient?
> 
> I don't feel strongly about it.

Using keg-only formulae often requires adding some extra paths to environment variables (see [this bit of code](https://github.com/ignition-tooling/release-tools/blob/cf14d7d8da831107a9f648b4e6f9fe0f5b36c564/jenkins-scripts/lib/project-default-devel-homebrew-amd64.bash#L113-L142) from the scripts for build.osrfoundation.org for an example). I was going to suggest the caveats printed by Homebrew as well, which can be viewed with `brew info`:

~~~
$ brew info qt@5
qt@5: stable 5.15.2 (bottled) [keg-only]
Cross-platform application and UI framework
https://www.qt.io/
...
==> Caveats
We agreed to the Qt open source license for you.
If this is unacceptable you should uninstall.

qt@5 is keg-only, which means it was not symlinked into /usr/local,
because this is an alternate version of another formula.

If you need to have qt@5 first in your PATH, run:
  echo 'export PATH="/usr/local/opt/qt@5/bin:$PATH"' >> /Users/scpeters/.bash_profile

For compilers to find qt@5 you may need to set:
  export LDFLAGS="-L/usr/local/opt/qt@5/lib"
  export CPPFLAGS="-I/usr/local/opt/qt@5/include"

For pkg-config to find qt@5 you may need to set:
  export PKG_CONFIG_PATH="/usr/local/opt/qt@5/lib/pkgconfig"

~~~

src/rosdep2/platforms/osx.py
```diff
@@ -204,7 +204,11 @@ def is_installed(r):
             pkg_info = pkg_info[0]
             linked_version = pkg_info['linked_keg']
             if not linked_version:
-                return False
+                keg_only = pkg_info['keg_only']
+                if keg_only:
+                    return True
```
scpeters, line 209: I think this code will `return True` for any keg-only formula, even if it's not actually installed. Do you want to use the `installed` field instead?

note that the `caveats` field in the json does not match the full output printed by `brew info`. I believe running `brew ruby -e 'require "caveats"; puts Caveats.new("qt@5".f).keg_only_text'` can generate the full caveats string

PR #802, opened 2021-03-31
Add support for a few RHEL clones

cottsay:
In addition to CentOS, this change adds support for a few other RHEL clones: AlmaLinux, Oracle Linux, and Rocky Linux.

### DIFF
--- a/src/rosdep2/platforms/redhat.py
+++ b/src/rosdep2/platforms/redhat.py
@@ -87,13 +87,13 @@ def register_rhel(context):
     context.set_os_version_type(OS_RHEL, lambda self: self.get_version().split('.', 1)[0])
 
 
-def register_rhel_clone(context, OS_RHEL_CLONE):
+def register_rhel_clone(context, os_rhel_clone_name):
     # Some distributions are rebuilds of RHEL and can be treated like RHEL
     # because they are versioned the same and contain the same packages.
     (os_name, os_version) = context.get_os_name_and_version()
-    if os_name == OS_RHEL_CLONE and not context.os_override:
+    if os_name == os_rhel_clone_name and not context.os_override:
         print('rosdep detected OS: [%s] aliasing it to: [%s]' %
-              (OS_RHEL_CLONE, OS_RHEL), file=sys.stderr)
+              (os_rhel_clone_name, OS_RHEL), file=sys.stderr)
         context.set_os_override(OS_RHEL, os_version.split('.', 1)[0])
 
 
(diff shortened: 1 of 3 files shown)


src/rosdep2/platforms/redhat.py
```diff
@@ -77,13 +87,13 @@ def register_rhel(context):
     context.set_os_version_type(OS_RHEL, lambda self: self.get_version().split('.', 1)[0])
 
 
-def register_centos(context):
-    # CentOS is an alias for RHEL. If CentOS is detected and it's not set as
-    # an override force RHEL.
+def register_rhel_clone(context, OS_RHEL_CLONE):
```
tfoote, line 90:
```suggestion
def register_rhel_clone(context, os_rhel_clone_name):
```

I started to flag this as failing to import the constant. But then I realized the all caps was just an argument. It's a lot more readable to make this lowercase so the reader knows it's not a constant.

src/rosdep2/platforms/redhat.py
```diff
@@ -77,13 +87,13 @@ def register_rhel(context):
...
+    # Some distributions are rebuilds of RHEL and can be treated like RHEL
+    # because they are versioned the same and contain the same packages.
     (os_name, os_version) = context.get_os_name_and_version()
-    if os_name == OS_CENTOS and not context.os_override:
+    if os_name == OS_RHEL_CLONE and not context.os_override:
```
tfoote, line 94:
```suggestion
    if os_name == os_rhel_clone_name and not context.os_override:
```

src/rosdep2/platforms/redhat.py
```diff
@@ -77,13 +87,13 @@ def register_rhel(context):
...
         print('rosdep detected OS: [%s] aliasing it to: [%s]' %
-              (OS_CENTOS, OS_RHEL), file=sys.stderr)
+              (OS_RHEL_CLONE, OS_RHEL), file=sys.stderr)
```
tfoote, line 96:
```suggestion
              (os_rhel_clone_name, OS_RHEL), file=sys.stderr)
```

tfoote (dismissed):
Looks good with one requested tweak. I'm find with a different name too if there's something clearer.

src/rosdep2/platforms/redhat.py
cottsay: dd7a48b

src/rosdep2/platforms/redhat.py
cottsay: dd7a48b

src/rosdep2/platforms/redhat.py
cottsay: This is a completely reasonable point.

Applied in dd7a48b.

wjwwood (approved):

PR #806, opened 2021-04-20
Add --no-test-deps option

yotabits:
The `--no-test-deps` option allows ignoring `test_depend` marked
dependencies. This option is useful in the case where software is deployed
using the install folder and rosdep is run on it for dependencies.

Please let me know if this looks reasonable.

### DIFF
--- a/src/rosdep2/installers.py
+++ b/src/rosdep2/installers.py
@@ -421,7 +421,8 @@ class RosdepInstaller(object):
         self.installer_context = installer_context
         self.lookup = lookup
 
-    def get_uninstalled(self, resources, implicit=False, verbose=False):
+    def get_uninstalled(self, resources, implicit=False, verbose=False,
+                        include_test_deps=True):
         """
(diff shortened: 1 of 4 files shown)


emersonknapp:
I would prefer an alternate approach. Instead of tying to this one specific dependency type, extend https://github.com/ros-infrastructure/rosdep/pull/789 - with a reverse option `--exclude-dependency-types` which would allow you to skip specifically `test`, or any other kind.

yotabits:
> I would prefer an alternate approach. Instead of tying to this one specific dependency type, extend #789 - with a reverse option `--exclude-dependency-types` which would allow you to skip specifically `test`, or any other kind.

So do I! I didn't saw that PR when I made mine, my bad! 
I will close it when #789 gets merged.

nuclearsandwich:
@yotabits I'm going to close this now that #789 is merged (with release coming shortly) if there's any changes in this PR that you think would still benefit rosdep please re-open or create a new PR. Thanks!

PR #808, opened 2021-04-28
Support slow/bad networks as well

kunaltyagi:
On some networks, the following error is observed:
```python
Hit https://raw.githubusercontent.com/ros/rosdistro/master/rosdep/python.yaml
ERROR: unable to process source [https://raw.githubusercontent.com/ros/rosdistro/master/rosdep/ruby.yaml]:
        <urlopen error ('_ssl.c:711: The handshake operation timed out',)> (https://raw.githubusercontent.com/ros/rosdistro/master/rosdep/ruby.yaml)
```

This PR resolves the issue by trying the connection multiple times. Increasing timeout was tried but did not always succeed.

Followup from #687 

@nuclearsandwich Would it be possible to have a test env for this?


src/rosdep2/sources_list.py
```diff
@@ -472,24 +484,48 @@ def update_sources_list(sources_list_dir=None, sources_cache_dir=None,
...
+    def err_handler(exception, retries_left, custom_handler=None, source=None):
+        if retries_left:
+            if debug:
+                print("Trying again. Retries left: {}".format(retries_left))
+        else:
+            if custom_handler is None:
+                raise exception
+            custom_handler(source, exception)
+
+    def retry(max_retries, functor, acceptable_exceptions=tuple(), error_handler=err_handler):
```
kunaltyagi, line 498: There's a retry module on pypi which can be used here

src/rosdep2/sources_list.py
```diff
@@ -501,16 +537,23 @@ def update_sources_list(sources_list_dir=None, sources_cache_dir=None,
...
-            if skip_eol_distros:
-                if distribution.get('distribution_status') == 'end-of-life':
-                    print('Skip end-of-life distro "%s"' % dist_name)
-                    continue
+                return
+        distribution = get_index().distributions[dist_name]
+        if skip_eol_distros:
+            if distribution.get('distribution_status') == 'end-of-life':
+                print('Skip end-of-life distro "%s"' % dist_name)
+                return
```
kunaltyagi, line 556: This is indented 1 level in the original commit, but I felt the indentation was wrong. PTAL, I can revert this if needed

kunaltyagi:
Failing tests (test_rosdep_sources_list):
* test_SourcesListLoader_create_default
* test_load_cached_sources_list

The failure messages look alright to me (bad host name not found). Will try reverting the whitespace change

PR #809, opened 2021-04-28
Require setuptools, src/rosdep2/platforms/pip.py imports pkg_resources

hrnciar:


cottsay (approved):
Here's the spot: https://github.com/ros-infrastructure/rosdep/blob/master/src/rosdep2/platforms/pip.py#L33

At the very least, this should be a soft dependency for the debians. That said, I don't think it would cause any harm to add it as a hard dependency despite pip not being used on all platforms.

nuclearsandwich (approved):
I made a note about the possibility of using python3-pkg-resources instead of python3-setuptools but I don't consider it a blocker.

PR #810, opened 2021-05-12
Add Conda (formerly: RoboStack) installer

Tobias-Fischer:
Hi there,

This PR adds an installer for [https://github.com/RoboStack/ros-noetic](RoboStack). This patch is already included in the conda-forge build of rosdep (see https://github.com/conda-forge/rosdep-feedstock). It works as intended on Linux, macOS and Windows, and has for example been used to install dependencies using rosdep in our proposed cross-platform CI for moveit (see https://github.com/ros-planning/moveit/pull/2604).

It requires https://github.com/ros-infrastructure/rospkg/pull/224 to be merged first.

/cc @wolfv @traversaro @nuclearsandwich

PR #811, opened 2021-05-12
Fixes for Windows

Tobias-Fischer:
This PR contains a minimal set of changes from https://github.com/ros-infrastructure/rosdep/pull/661 to get the RoboStack installer (https://github.com/ros-infrastructure/rosdep/pull/810) to work on Windows.

@nuclearsandwich already had a look at https://github.com/ros-infrastructure/rosdep/pull/661 and the last comment was to rework the Azure pipeline to use GitHub actions instead; in this PR I'm taking a more minimal approach and only pull in easy-to-review changes without touching tests etc. which we don't need for our RoboStack use-case.

The contents of the PR are already used within the conda-forge build of rosdep (see https://github.com/conda-forge/rosdep-feedstock). We confirmed that Windows works in our moveit PR to add cross-platform CI (https://github.com/ros-planning/moveit/pull/2604) which makes use of rosdep.

/cc @wolfv @traversaro

### DIFF
--- a/src/rosdep2/main.py
+++ b/src/rosdep2/main.py
@@ -153,7 +153,10 @@ ERROR: Rosdep cannot find all required resources to answer your query
     except UsageError as e:
         print(_usage, file=sys.stderr)
         print('ERROR: %s' % (str(e)), file=sys.stderr)
-        sys.exit(os.EX_USAGE)
+        if hasattr(os, 'EX_USAGE'):
+            sys.exit(os.EX_USAGE)
+        else:
+            sys.exit()
     except RosdepInternalError as e:
         print("""
 ERROR: Rosdep experienced an internal error.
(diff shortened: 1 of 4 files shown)


src/rosdep2/main.py
```diff
@@ -153,7 +153,10 @@ def rosdep_main(args=None):
     except UsageError as e:
         print(_usage, file=sys.stderr)
         print('ERROR: %s' % (str(e)), file=sys.stderr)
-        sys.exit(os.EX_USAGE)
+        if hasattr(os, 'EX_USAGE'):
+            sys.exit(os.EX_USAGE)
+        else:
+            sys.exit()
```
nuclearsandwich, line 159: Should this be a non-zero exit code to indicate an error even on platforms where EX_USAGE is not defined?

nuclearsandwich (commented):
One comment about the change in exit code on error. If I was certain the exit code should be non-zero I would make the change myself and merge the PR but I want to get feedback from @Tobias-Fischer to make sure there's not some reason the program should exit cleanly here.

src/rosdep2/main.py
Tobias-Fischer: Yes, I believe so. @traversaro - you are the Windows guru, is there a standard way of doing so or is simply "1" sufficient?

Tobias-Fischer: `os.EX_USAGE` returns 64 on Unix, so I decided to return the same non-zero value on Windows.

Tobias-Fischer:
All done now - would be great if you could have another look @nuclearsandwich

nuclearsandwich (approved):

nuclearsandwich:
Thanks @Tobias-Fischer!

PR #829, opened 2021-09-18
Drop support for EOL Ubuntu distros

cottsay:
https://wiki.ubuntu.com/Releases#List_of_releases

- Xenial:  Apr 2021
- Yakkety: Jul 2017
- Zesty:   Jan 2019
- Artful:  Jul 2018
- Cosmic:  Oct 2018
- Disco:   Apr 2019
- Eoan:    Oct 2019

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -10,9 +10,17 @@ jobs:
     build:
       strategy:
         matrix:
-          os: [ubuntu-18.04, macos-latest]
+          os: [ubuntu-18.04, ubuntu-20.04, macos-latest]
           python: [2.7, 3.6, 3.7, 3.8, 3.9]
           exclude:
+          - os: ubuntu-20.04
+            python: 2.7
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -10,21 +10,9 @@ jobs:
     build:
       strategy:
         matrix:
-          os: [ubuntu-16.04, ubuntu-18.04, macos-latest]
-          python: [2.7, 3.5, 3.6, 3.7, 3.8, 3.9]
+          os: [ubuntu-18.04, macos-latest]
```
nuclearsandwich, line 13: Does it make sense to add 20.04 CI as our test base for Python >= 3.8?

nuclearsandwich (approved):
I think we could optionally add ubuntu-20.04 to the test matrix to cover more recent versions of python but since we're using the setup-python action the system python is not in use so it's not perfect coverage of our binary distribution anyway.

PR #837, opened 2021-11-12
Compress HTTP with GZip where available

cottsay:
The server behind GitHub URLs supports GZip compressed content encoding, which we can leverage to compress the YAML database behind rosdep substantially.

This reduces the typical bandwidth usage of a call to `rosdep update` by 84%, from 410KiB to 64KiB. It doesn't appear to change the runtime performance on a moderately fast internet connection, but might be very helpful on less performant connections.

Note that there are a lot of other improvements we could make here, such as streaming the content from the URL directly through the GZip decompression to the YAML parser and using a context to close the connection instead of explicitly calling `.close()`. We should consider those improvements when we drop support for Python 2.

### DIFF
--- a/src/rosdep2/gbpdistro_support.py
+++ b/src/rosdep2/gbpdistro_support.py
@@ -1,7 +1,3 @@
-try:
-    from urllib.request import urlopen
-except ImportError:
-    from urllib2 import urlopen
 import yaml
 try:
     import urlparse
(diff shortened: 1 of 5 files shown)


nuclearsandwich (approved):
I'm always caught off guard when compression has to be enabled and handled rather than being the default. This is an easy enough handler to add and will be a huge benefit, especially for rosdep users who are far from the GitHub network or on constrained connections.

We've avoided additional dependencies like `requests` in ros-infrastructure packages in order to keep the dependency tree tight but someday we might look at https://urllib3.readthedocs.io For now though I still don't think it's worth the overhead.

PR #838, opened 2021-11-13
Add support for wildcard OS name

cottsay:
This is an implementation of my [proposed changes to REP 111](https://github.com/ros-infrastructure/rep/pull/329).

The motivations and behaviors behind this implementation are discussed there.

### DIFF
--- a/src/rosdep2/lookup.py
+++ b/src/rosdep2/lookup.py
@@ -99,12 +99,14 @@ class RosdepDefinition(object):
         """
         rosdep_key = self.rosdep_key
         data = self.data
+        queried_os = os_name
+        queried_ver = os_version
 
         if type(data) != dict:
             raise InvalidData('rosdep value for [%s] must be a dictionary' % (self.rosdep_key), origin=self.origin)
         if os_name not in data:
             if '*' not in data:
-                raise ResolutionError(rosdep_key, data, os_name, os_version, 'No definition of [%s] for OS [%s]' % (rosdep_key, os_name))
+                raise ResolutionError(rosdep_key, data, queried_os, queried_ver, 'No definition of [%s] for OS [%s]' % (rosdep_key, os_name))
             elif type(data['*']) != dict:
                 raise InvalidData('rosdep value under OS wildcard for [%s] must specify a package manager' % (rosdep_key))
             os_name = '*'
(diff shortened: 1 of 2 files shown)


src/rosdep2/lookup.py
```diff
@@ -103,7 +103,11 @@ def get_rule_for_platform(self, os_name, os_version, installer_keys, default_ins
         if type(data) != dict:
             raise InvalidData('rosdep value for [%s] must be a dictionary' % (self.rosdep_key), origin=self.origin)
         if os_name not in data:
-            raise ResolutionError(rosdep_key, data, os_name, os_version, 'No definition of [%s] for OS [%s]' % (rosdep_key, os_name))
+            if '*' not in data:
+                raise ResolutionError(rosdep_key, data, os_name, os_version, 'No definition of [%s] for OS [%s]' % (rosdep_key, os_name))
+            elif type(data['*']) != dict:
+                raise InvalidData('rosdep value under OS wildcard for [%s] must specify a package manager' % (rosdep_key))
+            os_name = '*'
```
tfoote, line 110: This is a little odd to set the query os_name to `*`. I see it all propagates through. However it will also propagate to the debug statements etc. I think it's slightly more complex but we should bookkeep the queried os and the evaluated os. Otherwise people will set an OS explicitly and then get errors saying that it can't resolve for `*` OS.

cottsay: I'm actually really glad you brought this up. I did this when I originally made the change, but opted for a simpler change to make the review easier. I'll preserve the queried OS name and version as suggested.

cottsay: 91cf3a1

test/test_rosdep_lookup.py
```diff
@@ -189,7 +189,7 @@ def test_RosdepDefinition():
     except ResolutionError as e:
         assert e.rosdep_key == 'trusty_only_key'
         assert e.os_name == 'ubuntu'
-        assert e.os_version == '*'
+        assert e.os_version == 'lucid'
```
cottsay, line 192: Note that this is a change in the existing behavior, but this pattern doesn't appear anywhere in the rosdep database right now.

nuclearsandwich: > this pattern doesn't appear anywhere in the rosdep database right now.

We don't have anything that fits this exactly but I think any situation where we use the combination of an os_version wildcard with specific `os_version: null` keys is meant to be covered by this test even if we don't have anywhere where a whole os_name is also null (as we usually just omit the key).

Am I right in thinking that this behavior change is due to https://github.com/ros-infrastructure/rosdep/pull/838/commits/91cf3a113e491877de5b3744a1102e616f7ba3d6 which preserved the originally requested os_name and os_version?
This definitely seems like an improvement to the error messages for which a case could be made independently from the os_name wildcard support so I see it as an improvement to existing behavior. But I think that the case under test is an

nuclearsandwich (approved):
Looks like a solid improvement!

One thing I did notice while reviewing which isn't a blocker for this PR is that the definition test is growing towards 160 lines of asserts all with varying state. It would be nice to break that up into smaller test methods at some point.

test/test_rosdep_lookup.py
cottsay: > Am I right in thinking that this behavior change is due to `91cf3a1` ...?

Correct.

However, if you follow the code, the `os_version` isn't getting changed until [L138](https://github.com/ros-infrastructure/rosdep/blob/be780c45312b562f461fd7d2f06232e4e66df1aa/src/rosdep2/lookup.py#L137-L138) at which point the only possible `ResolutionError` would be from an explicitly nulled-out rule. Combined, this means that there is only one pattern which will behave differently under 91cf3a1:
```yaml
foo:
  debian:
    '*': null
    sid: [bar]
```

This syntax is somewhat redundant, and causes rosdep's resolution error message to change, but not the exception type. When last I surveyed we had no occurrences of `'*': null` in the database, but it looks like two of them have slipped in. They should probably be removed.

An OS-level `null` will bypass L138 so the error message won't change. Again, though, we shouldn't have any of these in the database ([even though we do](https://github.com/ros/rosdistro/blob/d5bac54d9ca5485dfc5ac568bf3e743fa0e66282/rosdep/base.yaml#L90)).

tfoote (approved):

Issue #819, opened 2021-06-24
rosdep does not register succesful installation of custom dependencies through pip


Bart-van-Ingen:
Hi!

I want to use rosdep to install all dependencies for my ros workspace, to keep it as simple as possible. this includes a python package that I have made myself. I have been able to find the process of doing this as can be seen on my [ROS Answers question](https://answers.ros.org/question/380740/best-practice-installing-custom-dependencies-for-ros2/).

Together with gvdhoorn, I have come to the conclusion that rosdep is unable to successfully register the installation of the custom python project through pip.

The error that is indicate is:

``` bash
Installing collected packages: repo
Successfully installed repo-0.1
ERROR: the following rosdeps failed to install
pip: Failed to detect successful installation of [git+https://github.com/user/repo.git@master]
```

I changed the ```pip_detect()``` function in ```/usr/lib/python3/dist-packages/rosdep2/platforms/pip.py ``` by adding more prints to get more insight , as shown bellow, with the added prints indicated with ``` # <-----------``` 

``` python
def pip_detect(pkgs, exec_fn=None):
    print('pkgs =' + str(pkgs)) # <-----------
    """
    Given a list of package, return the list of installed packages.

    :param exec_fn: function to execute Popen and read stdout (for testing)
    """
    pip_cmd = get_pip_command()
    if not pip_cmd:
        return []

    fallback_to_pip_show = False
    print('exec_fn =' + str(exec_fn)) # <-----------
    if exec_fn is None:
        exec_fn = read_stdout
        fallback_to_pip_show = True
    pkg_list = exec_fn(pip_cmd + ['freeze']).split('\n')
    print('pkg_list = ' + str(pkg_list)) # <-----------

    ret_list = []
    for pkg in pkg_list:
        pkg_row = pkg.split('==')
        if pkg_row[0] in pkgs:
            ret_list.append(pkg_row[0])
    print('ret_list =' + str(ret_list)) # <-----------
    print('fallback_to_pip_show =' + str(fallback_to_pip_show)) # <-----------

    # Try to detect with the return code of `pip show`.
    # This can show the existance of things like `argparse` which
    # otherwise do not show up.
    # See:
    #   https://github.com/pypa/pip/issues/1570#issuecomment-71111030
    if fallback_to_pip_show:
        for pkg in [p for p in pkgs if p not in ret_list]:
            # does not see retcode but stdout for old pip to check if installed
            proc = subprocess.Popen(
[...]

PR #840, opened 2021-11-24
Fix for issue #819 - adding ability to deal with pip packages installed via git

suddrey-qut:
Suggested fix for issue #819


cottsay:
What is the desired behavior if a package is already installed from PyPI and a rosdep rule requests the package be installed from a specific git URL like this? From what I can tell, this change will leave the package alone.

Doom4535:
My 2 cents would be to have it prioritize the version from the specific URL and reinstall that fixed one with the assumption that it is an intentional development override; perhaps it should throw an alert/WARNING when doing so? Fixed URL packages do break traditional version tracking

PR #845, opened 2021-12-16
Add support for --sources-cache-dir to update verb

cottsay:
From what I can tell, support for this option was only implemented for the `install` verb.

From `rosdep update --help`:
```
  ...
  -c SOURCES_CACHE_DIR, --sources-cache-dir=SOURCES_CACHE_DIR
                        Override ~/.ros/rosdep/sources.cache
  ...
```

### DIFF
--- a/src/rosdep2/main.py
+++ b/src/rosdep2/main.py
@@ -651,7 +651,7 @@ def command_update(options):
         return 1
     try:
         print('reading in sources list data from %s' % (sources_list_dir))
-        sources_cache_dir = get_sources_cache_dir()
+        sources_cache_dir = options.sources_cache_dir
         try:
             if os.geteuid() == 0:
                 print("Warning: running 'rosdep update' as root is not recommended.", file=sys.stderr)
@@ -659,7 +659,9 @@ def command_update(options):
         except AttributeError:
             # nothing we wanna do under Windows
             pass
-        update_sources_list(success_handler=update_success_handler,
+        update_sources_list(sources_list_dir=sources_list_dir,
+                            sources_cache_dir=sources_cache_dir,
+                            success_handler=update_success_handler,
                             error_handler=update_error_handler,
                             skip_eol_distros=not options.include_eol_distros,
                             ros_distro=options.ros_distro)


nuclearsandwich:
I am torn between approving this as-is since it's a net improvement and concern that improving the piecemeal support for this option is going to increase its usage and create _more_ inconsistency.

Looking back it seems like this feature, introduced in 2c5cae353f53899cb24b1d6d44cfd23227cbeb12 is meant primarily (possibly exclusively) for testing.

looking for usage of `get_sources_cache_dir()` in the codebase turns up the following:


https://github.com/ros-infrastructure/rosdep/blob/98502586ddcdfb38503855c486f3561ef75b9d29/src/rosdep2/main.py#L303

This is the fetching of the sources list for use as the argument default so that's fine.

---

https://github.com/ros-infrastructure/rosdep/blob/98502586ddcdfb38503855c486f3561ef75b9d29/src/rosdep2/rospack.py#L52-L61

I haven't figured out if this is dead code or not but if it's at all usable right now it won't be affected by this patch without further work.

---

https://github.com/ros-infrastructure/rosdep/blob/98502586ddcdfb38503855c486f3561ef75b9d29/src/rosdep2/sources_list.py#L472-L473

This usage is the initialization of an optional kwarg `sources_cache_dir` for the [`update_sources_list`](https://github.com/ros-infrastructure/rosdep/blob/98502586ddcdfb38503855c486f3561ef75b9d29/src/rosdep2/sources_list.py#L449-L451) function.


https://github.com/ros-infrastructure/rosdep/blob/98502586ddcdfb38503855c486f3561ef75b9d29/src/rosdep2/sources_list.py#L554-L555

This usage is likewise the initialization of an optional kwarg `sources_cache_dir` for the [`load_cached_sources_list`](https://github.com/ros-infrastructure/rosdep/blob/98502586ddcdfb38503855c486f3561ef75b9d29/src/rosdep2/sources_list.py#L545) function.

---

All uses of the latter two functions _within rosdep2_ pass the optional kwarg. I think the smallest delta from this patch which would give me the confidence to approve it would be to make the kwarg required (or at least start throwing warnings up when no sources_cache_dir is provided)

If we just ignore rospack (`init_rospack_interface` is not called within rosdep2 at all) then I think this could go through.

If we wanted to do something more holistic like a refactoring which creates a single context object for the sources list directory, sources cache directory, and other similar parameters which should persist through all rosdep operations and use that ubiquitiously I could see that being a good direction to go as well.

As pointed out, it doesn't seem like this has _ever_ been implemented for update operations, which at least means that we've only ever written to a consistent incorrect default directory but 
without something to help assert that we would reliably be using the configured cache directory for all operations I think that starting to add support could just make an old problem new and potentially worse as we could be writing to inconsistently incorrect directories.

If this is all "a lot" and we want to punt, I think we can do an even smaller thing and just suppress the help for this flag and consider it just for testing ([as it is used in tests](https://github.com/ros-infrastructure/rosdep/blob/98502586ddcdfb38503855c486f3561ef75b9d29/test/test_rosdep_main.py#L114)).

peci1:
A more complete solution is provided in https://github.com/ros-infrastructure/rosdep/pull/908 . It was not only the sources cache, but also the meta cache that goes by default into `~/.ros/rosdep`. I've added options to configure both - either via CLI options, or via a single env variable `ROSDEP_CACHE_PATH`.

peci1:
I found three major references to `init_rospack_interface` on github:

1. https://github.com/ros-infrastructure/rospkg/blob/721a25c7da39837655f43daeae2820422341d215/src/rospkg/manifest.py#L414
2. https://github.com/ros/rospack/blob/ad85a874575bbed74124b722b42b545537cc6aa3/src/rospack.cpp#L1733
3. https://github.com/lagadic/robotpkg/blob/065f4f94160362c43659f9798dbdad4c491fc977/sysutils/py-rospkg/patches/patch-aa#L1-L16

ad 3) This only disables rosdep usage, so no problem.

ad 1 and 2) In both cases, the `ROSDEP_CACHE_PATH` environment variable introduced in #908 would work out of the box. Whereas the added options (`-c` from this PR, or both `-c` and `-m` from #908) would need more downstream work.

So I agree with @nuclearsandwich that the wisest option would probably be to hide `-c` and `-m` from the help message and keep them for usage in tests. The only user-facing interface should be the `ROSDEP_CACHE_PATH` environment variable.

Would that be acceptable?

PR #846, opened 2021-12-16
Improve consistency of stderr usage

cottsay:
For the most part, common error messages are already getting written to stderr. This change makes an effort to move the remaining errors and warnings to stderr as well.


nuclearsandwich (approved):
Big picture I think it would be nice to move to an output framework for logging output like bloom does so that we're standardizing / centralizing this info rather than having to remember sys.stderr everywhere but this is doubtless an improvement over the status quo.

I spent longer than is reasonable trying to grep through and find other missed instances of this and did not.

PR #847, opened 2021-12-20
Support RosdepLookup overrides for several commands

cottsay:
Some commands already supported a 'lookup' argument, so this change aligns with those.

For 'command_keys', the `RosdepLookup` argument was actually getting ignored entirely, which was almost certainly a bug.

### DIFF
--- a/src/rosdep2/main.py
+++ b/src/rosdep2/main.py
@@ -682,7 +682,8 @@ def command_update(options):
 
 
 def command_keys(lookup, packages, options):
-    lookup = _get_default_RosdepLookup(options)
+    if not lookup:
+        lookup = _get_default_RosdepLookup(options)
     rosdep_keys = get_keys(lookup, packages, options.recursive)
     prune_catkin_packages(rosdep_keys, options.verbose)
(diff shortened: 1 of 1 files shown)


src/rosdep2/main.py
```diff
@@ -682,7 +682,8 @@ def update_error_handler(data_source, exc):
 
 
 def command_keys(lookup, packages, options):
```
nuclearsandwich, line 684: It's unfortunate that lookup isn't an optional kwarg like the other command_ functions. I'm not sure it's worth an API-break to make them consistent but I could see adding a TODO or Issue to update it if we ever do an API review and major-bump.

nuclearsandwich (approved):

src/rosdep2/main.py
cottsay: There are a couple of "classes" of the command functions in this file. My first run at this change was to converge them into a single function signature, but it would have been tricky to maintain API backward compatibility, so I punted it.

In short, I agree. It would be nice to clean this up someday.

PR #849, opened 2022-01-06
split pip commands to support args

phil0stine:
This addresses the use-case where custom `pip` args need to be passed from a rosdep definition. This could be applied to other installers but is just done here for `pip`.

For example, a sample `rosdep.yaml` definition given below can be correctly processed and executed, whereas the current implementation will cause an error related to trying to pass 
`['sudo', '-H', 'pip2', 'install', '-U', 'torch-scatter -f https://data.pyg.org/whl/torch-1.8.1+cu111.html']`
to pip via subprocess. The correct subprocess call should be 
`['sudo', '-H', 'pip2', 'install', '-U', 'torch-scatter', '-f', 'https://data.pyg.org/whl/torch-1.8.1+cu111.html']`, 
which is achieved by this patch.
 

``` yaml
torch_scatter:
  ubuntu:
    bionic:
      pip: ["torch-scatter -f https://data.pyg.org/whl/torch-1.8.1+cu111.html"]
```

The error when trying to run this definition through the current implementation is :
```
executing command [sudo -H pip2 install -U torch-scatter -f https://data.pyg.org/whl/torch-1.8.1+cu111.html]
Invalid requirement: 'torch-scatter -f https://data.pyg.org/whl/torch-1.8.1+cu111.html'
It looks like a path. Does it exist ?    
```

### DIFF
--- a/src/rosdep2/platforms/pip.py
+++ b/src/rosdep2/platforms/pip.py
@@ -150,4 +150,5 @@ class PipInstaller(PackageManagerInstaller):
             cmd.append('-q')
         if reinstall:
             cmd.append('-I')
-        return [self.elevate_priv(cmd + [p]) for p in packages]
+
+        return [self.elevate_priv(cmd + p.split()) for p in packages]


cottsay:
From a technical perspective, you need to teach the [detect phase](https://github.com/ros-infrastructure/rosdep/blob/16e4b3a50f909dfff0f88e8b259cfe5ee308e7b1/src/rosdep2/platforms/pip.py#L80-L85) how to handle this as well, or rosdep will fail to detect successful installation right after doing it.

That said, a rule like this will never be in the official rosdep database because it violates the policies regarding authorized sources: https://github.com/ros/rosdistro/blob/master/CONTRIBUTING.md#pip

I'm not opposed to extending rosdep to support functions like this in private rosdep databases, but even then, showing command line arguments in package names seems like the wrong way to go about it. The database is supposed to contain package names, not installation commands.

phil0stine:
I thought something like this might be the response. Clearly just applying this to `pip` would not be the best approach but I figured it would start a conversation. Thanks for the feedback.

PR #850, opened 2022-01-23
Use unittest.mock where possible

cottsay:
The 'unittest.mock' module has been present since Python 3.3, and some platforms are now dropping the 'mock' backport package.

### DIFF
--- a/test/test_rosdep_alpine.py
+++ b/test/test_rosdep_alpine.py
@@ -29,7 +29,10 @@
 
 import os
 import traceback
-from mock import Mock, patch
+try:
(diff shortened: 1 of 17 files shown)


nuclearsandwich (approved):

nuclearsandwich:
In the spirit of the stairwell. Does it make sense to drop mock from the CI installation since all our tested platforms provide unittest.mock?

cottsay:
> Does it make sense to drop mock from the CI installation

From an offline discussion: Eventually, but it is still needed for Python 2.7, so it will need to be made conditional for now. I'll make that change in a follow-up PR.

PR #861, opened 2022-04-01
Pass with flake8_comprehensions

cottsay:
This flake8 plugin is installed as part of the ROS 2 development packages, so it's pretty likely that anyone developing rosdep will already have this plugin present. Since the plugin's mere presence will cause it to load and run, it's probably easiest to just make this package pass the checks.

They're valid issues, anyway.


.github/workflows/ci.yaml
```diff
@@ -31,7 +31,7 @@ jobs:
           python -m pip install --upgrade pip setuptools
           python -m pip install PyYAML argparse rospkg vcstools catkin_pkg python-dateutil rosdistro
           python -m pip install -e .
-          python -m pip install pytest pytest-cov flake8 mock codecov
+          python -m pip install pytest pytest-cov flake8 flake8-comprehensions mock codecov
```
nuclearsandwich, line 34: This package hasn't been updated to use extras_require but I'll open a PR for that once this is in to avoid further conflicts.

nuclearsandwich (approved):
Thanks for the improvements!

PR #862, opened 2022-04-01
Support PEP 338 invocation of rosdep2 module

cottsay:
This package has a well-established entry point. It would be great to gain access to that entry point in a way that allows use of a specific interpreter and bypasses PATH.

### DIFF
--- a/src/rosdep2/__main__.py
+++ b/src/rosdep2/__main__.py
@@ -1,0 +1,32 @@
+# Copyright (c) 2022, Open Source Robotics Foundation, Inc.
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+#     * Redistributions of source code must retain the above copyright
+#       notice, this list of conditions and the following disclaimer.
+#     * Redistributions in binary form must reproduce the above copyright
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
So I understand correctly, is goal of this PR is to enable the use of

`python3 -m rosdep2 update` so that it's equivalent (assuming your path is rational and the rosdep bin and python3 bin are from the same environment) to `rosdep update`

I don't doubt that this is the way to go about it but I'm not quite able to connect the dots myself between the `runpy.run_module` signature in https://peps.python.org/pep-0338/ and the addition of this file. So I'd appreciate some support in understanding how this works.

cottsay:
> I'm not quite able to connect the dots myself between the `runpy.run_module` signature in https://peps.python.org/pep-0338/ and the addition of this file.

From the abstract:

> This PEP defines semantics for executing any Python module as a script, either with the -m command line switch, or by invoking it via runpy.run_module(modulename).

I referenced this pep because it describes the "desired behavior" of how I'd like to invoke this module (both via `-m` and `runpy.run_module`). Unfortunately, I'm a little lost on the details of how the connection to `__main__` is made, but the non-pep documentation is pretty clear, and also references pep 338: https://docs.python.org/3/using/cmdline.html#cmdoption-m

PR #865, opened 2022-04-10
Add pip installer to arch linux platform.

russkel:
Added so Arch Linux can install deps via pip.


nuclearsandwich (approved):

nuclearsandwich:
I can't find a reference to it now, but I think that in the past there were difficulties because Arch Linux's pip defaulted to user install rather than system installation. Without references to back it up that's not something we should block the change on.

@russkel have you been using this patch and is it successfully installing system-wide pip packages?

russkel:
From what I could tell, running `pip` as sudo always installed system-wide for me, even if it already existed on the user's local libs.

PR #866, opened 2022-04-10
If Manjaro is detected override to Arch Linux

russkel:
I am trying to bring up a ros core install on my Manjaro machine and have been adding rosdep keys to `arch`. Manjaro uses the same packages as Arch Linux, only delaying updates a little while before pushing to manjaro. I have been using rosdep's override argument to install the deps, but making this PR seemed more appropriate.

There are no `manjaro` keys in rosdistro/rosdep/*.yml

### DIFF
--- a/src/rosdep2/platforms/arch.py
+++ b/src/rosdep2/platforms/arch.py
@@ -28,8 +28,9 @@
 # Author Tully Foote/tfoote@willowgarage.com
 
 import subprocess
+import sys
 
-from rospkg.os_detect import OS_ARCH
+from rospkg.os_detect import OS_ARCH, OS_MANJARO
 
 from ..installers import PackageManagerInstaller
 from .source import SOURCE_INSTALLER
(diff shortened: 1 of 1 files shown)


russkel:
Not entirely sure why the tests are failing now given that I copied existing code from `debian.py` to do this.

src/rosdep2/platforms/arch.py
```diff
@@ -28,8 +28,9 @@
 # Author Tully Foote/tfoote@willowgarage.com
 
 import subprocess
```
cottsay, line 30: > Not entirely sure why the tests are failing now given that I copied existing code from `debian.py` to do this.

This should resolve the issue:

```suggestion
from __future__ import print_function
import subprocess
```

russkel:
Ah - quite right.

nuclearsandwich:
If we're just considering "today" then this would be a pretty straightforward approval from me today. However, as I've heard @cottsay discuss slow separation of Raspbian from Debian I could see this getting rough in the same way.

Manjaro repositories are currently a superset of Arch Linux repositories... There are packages in Manjaro which are not in upstream Archlinux. I'm not too concerned with the "delay" between Arch availability and Manjaro available just the packages that are only in Manjaro.

@cottsay based on what you've observed with the Raspbian situation, do you think adding an override now would complicate a future transition to an independent Manjaro resolver?

cottsay:
> do you think adding an override now would complicate a future transition to an independent Manjaro resolver?

I don't think that an additional aliasing distro is going to complicate any future plans to "untangle" the aliases in general. Making Manjaro usable now seems like sufficient value to merge this now rather than wait for the resources for said untangling to manifest.

russkel:
My pleasure. Thanks for merging it in.

PR #867, opened 2022-04-11
Alias Raspbian to Debian

cottsay:
Debian is the upstream source for Debian packages.

`OS_RASPBIAN` was added to rospkg in ros-infrastructure/rospkg#244 and was released in 1.4.0.

Note that there aren't any detectors behind `OS_RASPBIAN` right now. It will be a breaking change for rospkg to switch detecting Raspbian as `debian` and start detecting as `raspbian` instead, and this change prepares for that eventuality.


nuclearsandwich (approved):
Thanks for remembering to update the minimum rospkg version.

PR #879, opened 2022-06-23
Bump minimum Python 3 version in stdeb to 3.6

cottsay:
Python 3.6 is the Python 3 version packaged for Bionic, and it is the minimum version we're running CI for.


nuclearsandwich (approved):
f-striiiiiiiiiiiiings :rocket:

cottsay:
> f-striiiiiiiiiiiiings

Too bad we still support Python 2 here...

nuclearsandwich:
> Too bad we still support Python 2 here...

FFFFFFFFFFFFFFFFFFFFFFFFFf, strings

PR #880, opened 2022-06-23
Update developer documentation to reflect mock dependency

cottsay:
Follow-up to #850


nuclearsandwich (approved):
Does this interact with https://github.com/ros-infrastructure/rosdep/pull/879 at all?

I don't think it's bad to have this information but is it now redundant?

cottsay:
> Does this interact with #879 at all?

Kinda, we still support Python 2.7, which needs `mock`. But no Python 3 versions will need it moving forward.

PR #882, opened 2022-06-23
Ignore rosdep's own deprecations when running rosdep tests

cottsay:
It isn't useful to see these warnings when running rosdep's tests.


nuclearsandwich (approved):

nuclearsandwich:
To satisfy my own curiosity, is this for RPM package generation compliance or just noticed during testing?

cottsay:
> ...is this for RPM package generation compliance or just noticed during testing?

I just noticed it while looking at the GitHub Actions logs. I'm sure it's in the RPM build logs as well, but it isn't causing any trouble there.

PR #884, opened 2022-06-23
Mark linter tests and tests which require network

cottsay:
Marking tests appropriately will allow us to be more selective about what tests are run.


nuclearsandwich (approved):
With the caveat that I didn't perform offline tests to see if these markers are comprehensive they do appear applied correctly where used. Thanks for the improvement!

PR #885, opened 2022-06-24
0.22.0

cottsay:


nuclearsandwich (approved):
Did some minor edits and rewords on the changelog.

PR #887, opened 2022-06-24
Drop ROS Python package dependencies in debs

### LINKED ISSUE #886: Rosdep broken after 0.22 release
We are experiencing CI failures across ros-planning organization on all ROS2 distros with the new rosdep version (0.22). It seems it cannot find / install the rospkg module. An example passing and failing CI runs are below respectively:

Pass: https://github.com/PickNikRobotics/rviz_visual_tools/actions/runs/2391349400/attempts/1
Fail: https://github.com/PickNikRobotics/rviz_visual_tools/runs/7046842854?check_suite_focus=true

Copying the error message we are getting below:

```
$ sudo rosdep init > /dev/null 
    File "/usr/bin/rosdep", line 6, in <module>
      from pkg_resources import load_entry_point
    File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 3267, in <module>
      def _initialize_master_working_set():
    File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 3[241](https://github.com/PickNikRobotics/rviz_visual_tools/runs/7046842854?check_suite_focus=true#step:5:246), in _call_aside
      f(*args, **kwargs)
    File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 3279, in _initialize_master_working_set
      working_set = WorkingSet._build_master()
    File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 573, in _build_master
      ws.require(__requires__)
    File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 891, in require
      needed = self.resolve(parse_requirements(requirements))
    File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 777, in resolve
      raise DistributionNotFound(req, requirers)
  pkg_resources.DistributionNotFound: The 'rospkg>=1.4.0' distribution was not found and is required by rosdep
  'sudo rosdep init > /dev/null ' returned with 1
```

I didn't go deep down to find out the root cause, just wanted to quickly notify folks on this.

cottsay:
When our Python packages are split into a separate `-modules` subpackage, it appears that the pattern is to retain the package distribution information in the main package, which contains the scripts.

When our packages take dependencies on only the `-modules` subpackage, they need to omit the Python dependencies to prevent `pkg_resources` from looking for those missing packages. Instead, we rely on apt and dpkg to ensure that the dependencies are present.

This shouldn't have any affect on non-debian installation of rosdep.

Fixes #886

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,7 +9,7 @@ kwargs = {
     'version': '0.22.0',
     'packages': ['rosdep2', 'rosdep2.ament_packages', 'rosdep2.platforms'],
     'package_dir': {'': 'src'},
-    'install_requires': ['catkin_pkg >= 0.4.0', 'rospkg >= 1.4.0', 'rosdistro >= 0.7.5', 'PyYAML >= 3.1', 'setuptools'],
+    'install_requires': ['PyYAML >= 3.1', 'setuptools'],
     'extras_require': {
         'test': [
             'flake8',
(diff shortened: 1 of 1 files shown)


cottsay:
Here's a similar pattern in `rosdistro`: https://github.com/ros-infrastructure/rosdistro/blob/aab3bf5ea6589691e352658a688a796c4d423b51/setup.py#L51-L58

clalancette (approved):

PR #892, opened 2022-09-07
Support for URLs with basic authentication

### LINKED ISSUE #891: Interest in development of support for authentication in rosdep
Apt has support for authentication for private APT repository servers.
I was wondering if I were to build such (or similar) functionality for rosdep, would this be something that the maintainers accept in a pull request or is this something that is beyond the scope of OSRF/rosdep?

AFAIK only authentication for proxies is supported which is different from what APT offers.

cottsay:
This change adds support for basic HTTP authentication values in URLs.

### DIFF
--- a/src/rosdep2/url_utils.py
+++ b/src/rosdep2/url_utils.py
@@ -25,16 +25,23 @@
 # ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
 
+import base64
 from gzip import GzipFile
 from io import BytesIO
 try:
     from urllib.request import urlopen
     from urllib.error import URLError
+    from urllib.parse import splituser
+    from urllib.parse import urlparse
+    from urllib.parse import urlunparse
     import urllib.request as request
 except ImportError:
+    from urllib2 import splituser
     from urllib2 import urlopen
     from urllib2 import URLError
     import urllib2 as request
+    from urlparse import urlparse
+    from urlparse import urlunparse
 
 from ._version import __version__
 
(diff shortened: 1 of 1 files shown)


cottsay:
I have mixed feelings about the utility of this change as well, but I also believe it is a simple improvement. Another reason I'm not super happy with this change is that the password is printed to stdout whenever an update is performed. I looked into printing the "sanitized" URL instead, but the best I could do was to repeat the same process in another spot, so it felt forced. Maybe that can be addressed in a follow-up change.

There are some tricks you can utilize to make the passwords less universally readable. You could forego the global rosdep sources list and protect the sources list that contains secrets under your home directory:
```
$ mkdir -p ~/.ros/rosdep/sources.list.d
$ chmod o-rwx ~/.ros/rosdep/sources.list.d
$ export ROSDEP_SOURCE_PATH=~/.ros/rosdep/sources.list.d
$ rosdep init  # Note: no sudo
$ rosdep update
```
After that, export the `ROSDEP_SOURCE_PATH` variable as above in your `.bashrc` and add the secret URL to a new file in `~/.ros/rosdep/sources.list.d/`. You should be able to continue using `rosdep` and `bloom` as usual after that.

cottsay:
> I was thinking of utilizing the gnome-keyring for storing the credentials on a per-user basis, what do you think?

I'm very hesitant to take an unconditional dependency on any gnome libraries just to implement that feature, but I agree that it would be ideal. Maybe there is a lightweight library that can do it opportunistically, but keep in mind that we're still supporting Python 2 in rosdep and many newer libraries don't support that anymore.

progtologist:
I will give it a shot in the next days and see what I can come up with, I would also prefer a cleaner and more streamlined solution, but not at the expense of backwards compatibility.
Let me see what I can come up with, worst case scenario, we fallback to your implementation!

Issue #900, opened 2022-10-05
rosdep unable to determine installation of pip packages with extras or versions


JWhitleyWork:
I don't yet have a fix for this but just wanted to document the findings. When specifying Python packages to be installed by `pip` that include either a specific [version](https://peps.python.org/pep-0508/#versions) (e.g. `package==1.0.0`) or with [extras](https://peps.python.org/pep-0508/#extras) (e.g. `package[extra]`), `pip` is able to install the packages correctly but `rosdep` is unable to confirm installation of the packages. More complete example:

Snippet of custom YAML file:
```
python3-mycroft-plugin-tts-mimic3-pip:
  debian:
    pip:
      packages: ["mycroft-plugin-tts-mimic3[all]"]
  ubuntu:
    pip:
      packages: ["mycroft-plugin-tts-mimic3[all]"]
...
python3-pyee8-pip:
  debian:
    pip:
      packages: ["pyee==8.1.0"]
  ubuntu:
    pip:
      packages: ["pyee==8.1.0"]
```

Output of `rosdep` when installing these dependencies:

```
Installing collected packages: mycroft-plugin-tts-mimic3
Successfully installed mycroft-plugin-tts-mimic3-0.1.5
WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv
executing command [sudo -H pip3 install -U pyee==8.1.0]
Collecting pyee==8.1.0
  Using cached pyee-8.1.0-py2.py3-none-any.whl (12 kB)
Installing collected packages: pyee
Successfully installed pyee-8.1.0
WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv
ERROR: the following rosdeps failed to install
  pip: Failed to detect successful installation of [mycroft-plugin-tts-mimic3[all]]
  pip: Failed to detect successful installation of [pyee==8.1.0]
```

vonZeppelin:
Bumped into this too. I think there is yet another issue that can be tracked under this ticket: `rosdep` fails to detect successful installation of pip packages specified as [VCS links](https://pip.pypa.io/en/stable/topics/vcs-support/), at least the error message at the end of installation is the same.

[...]

PR #901, opened 2022-10-06
Handle pip package versions like pip does

JWhitleyWork:
With most ROS-supported operating systems and their default package managers, major version changes will usually not be automatically introduced. This avoids major incompatibilities being introduced within a given, supported operating system version for packages installed through the default package manager. As PyPi retains Python modules which are not tied to operating system versions, major, breaking changes can be introduced unexpectedly. Therefore, there exists a need for `rosdep` to be able to handle version management of `pip`-installed modules. This PR introduces version checking to `rosdep` when verifying pip package installation.

Since `rosdep` passes package requirements directly from a keys YAML file to `pip` without modification, the installation of packages with version requirements was already possible. However, if a package with a version requirement was specified and installed, `rosdep` did not understand how to verify that the version requirement had been met. This PR adds this functionality.

Since this PR uses the [`packaging`](https://pypi.org/project/packaging/) module, it automatically complies with https://peps.python.org/pep-0440/ and https://peps.python.org/pep-0508/.

Covers the version-checking requirement (but not the extras-checkig one) of #900.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,6 +50,6 @@ elif 'SKIP_PYTHON_SCRIPTS' in os.environ:
     kwargs['name'] += '_modules'
     kwargs['entry_points'] = {}
 else:
-    kwargs['install_requires'] += ['catkin_pkg >= 0.4.0', 'rospkg >= 1.4.0', 'rosdistro >= 0.7.5']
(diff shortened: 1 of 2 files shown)


JWhitleyWork:
This ended up generating a few unexpected issues so I'm closing this PR. I'll keep the fork/branch in case anyone wants to try to complete the implementation in the future but we founmd other work-arounds for our use-case.

peci1:
Continuation in #909. @JWhitleyWork could you please chime in and leave some comments on the things #901 broke for you?

PR #907, opened 2022-11-22
feat(main): --packages-select and --packages-skip

Guillaumebeuzeboc:
Add options to only process or skip packages like
colcon can do.
Rosdep already support COLCON_IGNORE and CATKIN_IGNORE but with these flags no need to manually place those files.


artivis:
Hi @cottsay, could we get an approval for the ci workflow to run please?

PR #908, opened 2023-01-02
Support --sources-cache-dir option for the update command.

peci1:
Currently, the sources cache location cannot be configured when calling `rosdep update` from command line. This could be helpful e.g. on the buildfarm or in CI scripts - it would then be possible to completely isolate the build from the host environment and filesystem - `ROSDEP_SOURCE_PATH` would define the sources list path for both `init` and `update` commands, and `--sources-cache-dir` would then specify the cache path for `update`.

The `install` verb already respects the `--sources-cache-dir` option, so it would be possible to call the `install` command against the custom cache location created by the `update` command.

I haven't figured out a way to write a test for the changes, as there is no existing test for the `update` verb. But I hope the changes are simple enough to go through without tests. Locally, I tested the `update` verb both with and without the argument and it behaves as expected.

### DIFF
--- a/src/rosdep2/main.py
+++ b/src/rosdep2/main.py
@@ -653,7 +653,7 @@ def command_update(options):
     try:
         if not options.quiet:
             print('reading in sources list data from %s' % (sources_list_dir))
-        sources_cache_dir = get_sources_cache_dir()
+        sources_cache_dir = options.sources_cache_dir if options.sources_cache_dir else get_sources_cache_dir()
         try:
             if os.geteuid() == 0:
                 print("Warning: running 'rosdep update' as root is not recommended.", file=sys.stderr)
@@ -661,7 +661,8 @@ def command_update(options):
         except AttributeError:
             # nothing we wanna do under Windows
             pass
-        update_sources_list(success_handler=update_success_handler,
+        update_sources_list(sources_cache_dir=sources_cache_dir,
+                            success_handler=update_success_handler,
                             error_handler=update_error_handler,
                             skip_eol_distros=not options.include_eol_distros,
                             ros_distro=options.ros_distro,


peci1:
This PR would allow to use rosdep cache during tests on the ROS buildfarm while maintaining minimal interference with the host system: https://github.com/ros-infrastructure/ros_buildfarm/issues/923 .

nuclearsandwich:
Thank you for the detailed PR writeup! A cursory review of this PR shows me that it's very thorough.

I particularly appreciate the testing methodology, setting .ros to read-only in order to reveal off-spec cache usage is very promising. I'm going to bring this PR and my initial review to the next infrastructure team meeting to discuss and do a group review.

peci1:
Thanks for looking at this and for bringing it to the team meeting. Let me know if there is anything else I could do in this regard.

peci1:
> I'm going to bring this PR and my initial review to the next infrastructure team meeting to discuss and do a group review.

Has the meeting already been held? :)

peci1:
Friendly ping.

doc/overview.rst
```diff
@@ -29,13 +29,30 @@ rosdep gets its data from.
 ``rosdep update`` reads through this sources list to initialize your
 local database.
 
+You can override the location by setting environment variable
+``ROSDEP_SOURCE_PATH``. The custom path has to exist prior to calling
+``rosdep init``, otherwise the default one will be used.
```
nuclearsandwich, line 34: I would prefer that an "invalid" custom path produce an error rather than falling back to the default since that's probably not what the user intends if they're setting this value in the first place.

doc/overview.rst
```diff
@@ -29,13 +29,30 @@ rosdep gets its data from.
...
+
+Please note that when using ``sudo``, environment
+variables from the user are not passed to the command. To specify the variable
+for initializing the database, call::
+
+    sudo mkdir -p /usr/rosdep.sources
+    sudo env ROSDEP_SOURCE_PATH=/usr/rosdep.sources rosdep init
```
nuclearsandwich, line 41: `--preserve-env=` is another method for passing environment context to sudo commands.

```suggestion
Please note that when using ``sudo``, environment
variables from the user are not passed to the command. To specify the variable
for initializing the database, call::

    sudo mkdir -p /usr/rosdep.sources
    sudo env ROSDEP_SOURCE_PATH=/usr/rosdep.sources rosdep init
    
Alternatively you can use `--preserve-env` to pass through current environment variable values.
    export ROSDEP_SOURCE_PATH=/tmp/example-rosdep-sources
    mkdir $ROSDEP_SOURCE_PATH
    sudo --preserve-env=ROSDEP_SOURCE_PATH rosdep init
```

doc/overview.rst
```diff
@@ -95,5 +112,9 @@ dependencies::
```
nuclearsandwich, line 114: Do we need three line breaks between the content above and this note?

doc/sources_list.rst
```diff
@@ -30,6 +30,26 @@ Sources list file format
...
+    sudo env ROSDEP_SOURCE_PATH=/usr/rosdep.sources rosdep init
+
+If you point ``ROSDEP_SOURCE_PATH`` to a user-writable directory, you can avoid
+obtaining root privileges, so you can just call::
+
+    mkdir -p $HOME/my_rosdep_sources
+    ROSDEP_SOURCE_PATH=$HOME/my_rosdep_sources rosdep init
+
+Be aware that ``~`` is not expanded in ``ROSDEP_SOURCE_PATH``, so you should
+specify an absolute path.
```
nuclearsandwich, line 51: Thinking aloud: I pondered whether it was worth mentioning that some shells will expand `~` before setting the value in the environment variable. It's probably better to leave that off and encourage the use of `$HOME` so this is good as is.

doc/overview.rst
```diff
@@ -29,13 +29,30 @@ rosdep gets its data from.
...
 You can update your rosdep database by running::
 
     rosdep update
 
+If you have specified a custom ``ROSDEP_SOURCE_PATH``, do not forget to set it
+also for this command.
+
+Default location of the local rosdep database is in ``$HOME/.ros/rosdep``.
+To change it, set environment variable ``ROSDEP_CACHE_PATH``, or pass
+command-line arguments ``--sources-cache-dir`` and ``--meta-cache-dir``.
```
nuclearsandwich, line 55: Which value is used in the event both `ROSDEP_CACHE_PATH` and `--sources-cache-dir` are used? I would normally expect command line arguments to override environment variables but this PR has 'deprecate `--sources-cache-dir` in the title.

Can you please document what happens in this case?

nuclearsandwich (changes requested):
I focused my review on the documentation rather than implementation.

Most comments are small formatting questions or suggestions which @peci1 may accept, modify, or decline at their discretion.

There is one documented behavior that I think should be changed, which I've described inline. As well as a request to document how precedence between the environment variables and command line flags should be handled.

Thanks for taking on this large feature!

doc/sources_list.rst
peci1: I prefer keeping things simple for the average user. Hackers will know what they're doing and when it will work.

doc/overview.rst
peci1: Documented, CLI options win.

doc/overview.rst
peci1: This is a behavior this PR hasn't touched. It might seem a bit unfortunate, but I fear changing it, as this behavior has been there for 9 years already.

The env var behaves as a path list. If at least one path from it exists, it is used. If it is empty or all paths do not exist, the system default is used.

peci1:
@nuclearsandwich I followed your review and hopefully fixed all issues.

I needed to integrate #929 into this PR to make some tests succeed.

Could you please allow the workflow on this PR to get the buildfarm tests?

PR #909, opened 2023-01-12
Handle pip package versions like pip does [rev2]

peci1:
Continuation of #901.

This would come handy e.g. with package m2r added in https://github.com/ros/rosdistro/pull/35827 , where the newest released version on pip is python3-only, but it is also offered on python2 pip on bionic.

### DIFF
--- a/src/rosdep2/platforms/pip.py
+++ b/src/rosdep2/platforms/pip.py
@@ -34,6 +34,8 @@ import pkg_resources
 import subprocess
 import sys
 
+from packaging.requirements import Requirement
+from packaging.version import Version
 from ..core import InstallFailed
 from ..installers import PackageManagerInstaller
 from ..shell_utils import read_stdout
@@ -79,7 +81,8 @@ def is_cmd_available(cmd):
 
 def pip_detect(pkgs, exec_fn=None):
     """
-    Given a list of package, return the list of installed packages.
+    Given a list of package specifications, return the list of installed
+    packages which meet the specifications.
 
     :param exec_fn: function to execute Popen and read stdout (for testing)
     """
@@ -94,10 +97,20 @@ def pip_detect(pkgs, exec_fn=None):
     pkg_list = exec_fn(pip_cmd + ['freeze']).split('\n')
 
     ret_list = []
+    version_list = []
+    req_list = []
+
     for pkg in pkg_list:
         pkg_row = pkg.split('==')
-        if pkg_row[0] in pkgs:
-            ret_list.append(pkg_row[0])
+        version_list.append((pkg_row[0], Version(pkg_row[1])))
+
+    for pkg in pkgs:
+        req_list.append(Requirement(pkg))
+
+    for req in req_list:
+        for pkg in [ver for ver in version_list if ver[0] == req.name]:
+            if pkg[1] in req.specifier:
+                ret_list.append(req.name)
 
     # Try to detect with the return code of `pip show`.
     # This can show the existance of things like `argparse` which
@@ -105,18 +118,25 @@ def pip_detect(pkgs, exec_fn=None):
     # See:
     #   https://github.com/pypa/pip/issues/1570#issuecomment-71111030
     if fallback_to_pip_show:
-        for pkg in [p for p in pkgs if p not in ret_list]:
+        for req in [r for r in req_list if r.name not in ret_list]:
             # does not see retcode but stdout for old pip to check if installed
             proc = subprocess.Popen(
-                pip_cmd + ['show', pkg],
+                pip_cmd + ['show', req.name],
                 stdout=subprocess.PIPE,
                 stderr=subprocess.STDOUT
             )
             output, _ = proc.communicate()
             output = output.strip()
             if proc.returncode == 0 and output:
-                # `pip show` detected it, add it to the list.
-                ret_list.append(pkg)
+                # `pip show` detected it, check the version.
+                show_split = output.split('\n')
+
+                for line in [l for l in show_split if l.startswith('Version:')]:
+                    version = Version(line.strip().split()[1])
+
+                    if version in req.specifier:
+                        # version matches, add it to the list
+                        ret_list.append(req.name)
 
     return ret_list
 
--- a/src/rosdep2/platforms/pip.py
+++ b/src/rosdep2/platforms/pip.py
@@ -108,7 +108,7 @@ def pip_detect(pkgs, exec_fn=None):
     if exec_fn is None:
         exec_fn = read_stdout
         fallback_to_pip_show = True
-    pkg_list = exec_fn(pip_cmd + ['freeze']).split('\n')
+    pkg_list = exec_fn(pip_cmd + ['list', '--format', 'freeze']).split('\n')
     pkg_list = [p for p in pkg_list if len(p) > 0]
 
     ret_list = []
--- a/src/rosdep2/platforms/pip.py
+++ b/src/rosdep2/platforms/pip.py
@@ -102,6 +102,10 @@ def pip_detect(pkgs, exec_fn=None):
 
     for pkg in pkg_list:
         pkg_row = pkg.split('==')
+        
+        # Account for some unusual instances of === instead of ==
+        pkg_row[1] = pkg_row[1].strip('=')
+
         version_list.append((pkg_row[0], Version(pkg_row[1])))
 
     for pkg in pkgs:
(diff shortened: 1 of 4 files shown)


peci1:
Thanks @danielcranston , I've integrated your changes.

fujitatomoya (commented):
@tfoote @sloretz @cottsay 

either of you could take a look at this, related to https://github.com/ros2/ros2_documentation/pull/6187

this enables rosdep to properly handle pip package version specifiers (like package==1.2.3 or package>=1.0). Currently, rosdep doesn't "truly" support pip versioning. if you specify a version in the package name, rosdep can't detect that the versioned package is already installed and will try to reinstall it every time.

Note: this PR came up in ROS APAC Developer Meeting

stdeb.cfg
```diff
@@ -11,7 +11,7 @@ X-Python3-Version: >= 3.6
 Setup-Env-Vars: SKIP_PYTHON_MODULES=1
 
 [rosdep_modules]
-Depends3: ca-certificates, python3-rospkg-modules (>= 1.4.0), python3-yaml, python3-catkin-pkg-modules (>= 0.4.0), python3-rosdistro-modules (>= 0.7.5), python3 (>= 3.8) | python3-importlib-metadata, sudo
+Depends3: ca-certificates, python3-rospkg-modules (>= 1.4.0), python3-yaml, python3-catkin-pkg-modules (>= 0.4.0), python3-rosdistro-modules (>= 0.7.5), python3 (>= 3.8), python-packaging | python3-importlib-metadata, sudo
```
cottsay:
```suggestion
Depends3: ca-certificates, python3-rospkg-modules (>= 1.4.0), python3-yaml, python3-catkin-pkg-modules (>= 0.4.0), python3-rosdistro-modules (>= 0.7.5), python3 (>= 3.8) | python3-importlib-metadata, python3-packaging, sudo
```

peci1: Thanks for catching this.

cottsay:
Taking a broader view of our package management ecosystem, the main reason I hesitate to support version data in the rosdep database is that we already have version information on dependencies as part of `package.xml` files. Each declared dependency [may specify version requirements](https://www.ros.org/reps/rep-0149.html#dependency-tags) for that specific dependency.

It's worth calling out that rosdep (as a tool) doesn't do anything interesting with this version information right now. When the rosdep database is used by Bloom, it translates the version requirements to the system package metadata format so the values are used, just not here in rosdep. So debian/rpm rules already have and use version metadata coming from the package manifests and not the rosdep database, and I see this as precedent.

IMO, the only reasons that rosdep (the tool) doesn't use the manifest's version constraints today stem from:
1. Nobody has been motivated enough to champion it. Maybe that has now changed.
2. Ambiguity on how to properly handle a situation where a package is present but not at a compatible version. This is really just about establishing a policy or playbook for how we expect the underlying package managers to act.
3. Ambiguity on how to properly handle a situation where one operation calls for an impossible combination of version numbers.
4. Inability to make the underlying package managers do what we want. For example, I'm not sure it's possible to ask `apt` to perform an installation operation with a version range or even an explicit version that doesn't also contain the debian-specific parts of the version.

peci1:
I understand your concerns and I agree that enforcing versions on APT is almost impossible (because most public repos like Debian/Ubuntu only keep the last version).

But package (non-)managers like pip are a completely different story. They usually keep all versions, so it is easy to ask for any version you like. Please note the initial motivation for this change, which was a pip package whose latest version on pypi stopped supporting Python 2 but some older version did support it.

Without being able to specify version constraints, this package could just be thrown away for Bionic because the latest version is just not compatible. Or it could be left in rosdep, but that would be even worse because it would lead to impossible install instructions for pip (or to broken package, I don't remember which one).

Given the fact that pip packages can be updated any time, it can quite easily happen that some normally working rosdep key will stop working for some platform at any time (e.g. require a too new Python version).

This is why I think pip packages should have the possibility of having version constraints in rosdep, too. The ones from package.xml are not used for anything for Python/pip packages.

Maybe the solution could be to add a parameter to the installers that would tell whether version constraints are supported? Or maybe even directly a version constraint validation function (which could just return False on APT).

danielcranston:
Thanks for reviving this conversation!

> [...] I hesitate to support version data in the [official] rosdep database [because ...]

I agree with those reasons, and for clarity I wouldn't equate this PR to any implication that we should start versioning the pip packages in the official rosdep database.

This PR just extends rosdep (the tool) to _support_ versioned pip packages, where the only place you'd probably ever see them (for now, or maybe forever) would be in custom defined rosdep rules.

> [!Note]
> Just to clarify the value of being able to do this: it unlocks the possibility
> * To only rely on rosdep to install the dependencies of your project [which requires very specific pip package versions]
>   - (no conda/venvs/etc to force/pin those pip versions)
> * To specify all your package dependencies in your `package.xml` files
>   - (no requirements.txt, no manual `pip install` invocations alongside `rosdep install`)

At the same time I understand this is a bit off the beaten path of regular rosdep usage, and it's somewhat in contention with the `package.xml` spec since this adds a separate way to specify version (beyond the `package.xml` spec way. But again, rosdep currently is not using `package.xml`-declared version requirements and doing so requires answering @cottsay's bullet points above).

---

@cottsay (and other maintainers) do you mind sharing your thoughts on whether the "value" explained in the Note above is enough to warrant moving forward with this PR as-is? I'm expecting "no" 😄 (which is fine, I just want to confirm that you see the value proposition)

As for tackling your bullet points, I'm definitely interested in helping out, though I think there's some things I'd need to brush up on.

src/rosdep2/platforms/pip.py
```diff
@@ -146,32 +163,54 @@ def pip_detect(pkgs, exec_fn=None):
...
             if proc.returncode == 0 and output:
-                # `pip show` detected it, add it to the list.
-                ret_list.append(pkg)
+                # `pip show` detected it, check the version.
+                show_split = output.split('\n')
+
+                for line in [s for s in show_split if s.startswith('Version:')]:
+                    version = parse_version(line.strip().split()[1])
+
+                    if version is None or version in req.specifier:
```
cottsay, line 211: I had to dig into the `packaging` API to understand this, but the `specifier` object implements `__contains__` to make this syntax work. It behaves similar to the `range` type. :+1:

src/rosdep2/platforms/pip.py
```diff
@@ -146,32 +163,54 @@ def pip_detect(pkgs, exec_fn=None):
     if exec_fn is None:
         exec_fn = read_stdout
         fallback_to_pip_show = True
-    pkg_list = exec_fn(pip_cmd + ['freeze']).split('\n')
+    pkg_list = exec_fn(pip_cmd + ['list', '--format', 'freeze']).split('\n')
```
cottsay, line 166: The output of these commands is very similar. What advantage is there to using `pip list` over `pip freeze`?

Does this change the minimum version of pip required?

src/rosdep2/platforms/pip.py
```diff
@@ -132,9 +134,24 @@ def is_cmd_available(cmd):
...
+        return parse(version_str)
+    except InvalidVersion:
+        return None
+
+
 def pip_detect(pkgs, exec_fn=None):
     """
-    Given a list of package, return the list of installed packages.
+    Given a list of package specifications, return the list of installed
+    packages which meet the specifications.
```
cottsay, line 154: I'm not sure that returning just the names is the right move here. For example, if `pkgs` contains mutually exclusive specifiers (for example, `empy<4` and `empy>=4`), the function will simply return `['empy']` and rosdep will report success, even though only one of the specifiers was actually detected.

I'm not sure what the "right" behavior is here, but I don't think that rosdep should indicate success.

cottsay (changes requested):
> ...do you mind sharing your thoughts on whether the "value" explained in the Note above is enough to warrant moving forward with this PR as-is?

For clarity, I'm still opposed to including any version information in the official rosdep database, but I've been sufficiently convinced that this change provides value.

src/rosdep2/platforms/pip.py
peci1: It works on pip 20.0.2 from Ubuntu 20.04/Python 3.8. Do we need to support even older?

src/rosdep2/platforms/pip.py
peci1: What if this function returned a dict instead where each pkg would either have a value saying which pip package satisfies it, or None, which would mean it is not satisfied?

Or, maybe a better and simpler solution: what if we checked `pkgs` for conflicts explicitly? If there is a conflict like in your example, I think rosdep should fail anyways, shouldn't it?

peci1 (commented):
@cottsay I've responded to your inline questions.

PR #919, opened 2023-04-24
add openeuler support

Z572:
require https://github.com/ros-infrastructure/rospkg/pull/260
https://github.com/ros/rosdistro/pull/36933

### DIFF
--- a/src/rosdep2/platforms/redhat.py
+++ b/src/rosdep2/platforms/redhat.py
@@ -37,7 +37,8 @@ from rospkg.os_detect import (
     OS_FEDORA,
     OS_ORACLE,
     OS_RHEL,
-    OS_ROCKY
+    OS_ROCKY,
+    OS_OPENEULER
 )
 
 from .pip import PIP_INSTALLER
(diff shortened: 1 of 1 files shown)


tfoote (commented):
Overall we're happy to add support for additional OSs. However as I haven't heard of this one. Can you please provide a little bit more context as to the user base and the package management system and database. In particular I see that it uses the DNF and YUM installers. Does it track/follow/use one of the other package databases or does it have it's own package database for example we have a lot of rhel clones that don't need separate database entries? This will be important that we have a set of rules .

Z572:
openEuler (formerly EulerOS) is a Red-Hat-like Linux distribution. It is the successor of Huawei's EulerOS.

openEuler is an open source project incubated and operated by the OpenAtom Foundation.

It use DNF as package-management, but no base on other distribution, many package name are different from rhel, like RHEL named "libzstd-devel",but openEuler named "zstd-dev"

have some linux Commercial Editions base on it, see https://www.openeuler.org/en/download/commercial-release/

official website: https://www.openeuler.org/en/
use case: https://www.openeuler.org/en/showcase/
you can search package from https://pkgmanage.openeuler.org/Packagemanagement or https://oepkgs.net/

wuxiaoqiang12:
@tfoote Could you please review and accept this PR? As @Z572 has already elaborated on the openEuler OS context, and the rospkg side has merged the relevant PR. Now the rosdistro part is pending your review, as noted in the comment of [rosdistro/pull/36933](https://github.com/ros/rosdistro/pull/36933)

tfoote:
Sorry I'm having trouble submitting the review comments formally. Thanks for the extra info and looking at 24.09 it looks like there's been a big push since this was submitted.

Reviewing the implementation, can you please register it in alphabetical order and have the definition in the same order for easier readability maintenance in the future.

src/rosdep2/platforms/redhat.py
```diff
@@ -61,14 +62,22 @@ def register_installers(context):
 def register_platforms(context):
     register_fedora(context)
     register_rhel(context)
-
+    register_openeuler(context)
```
tfoote, line 65: Please move this up to be in alphabetical order.

src/rosdep2/platforms/redhat.py
```diff
@@ -61,14 +62,22 @@ def register_installers(context):
...
     # Aliases
     register_rhel_clone(context, OS_ALMALINUX)
     register_rhel_clone(context, OS_CENTOS)
     register_rhel_clone(context, OS_ORACLE)
     register_rhel_clone(context, OS_ROCKY)
 
 
+def register_openeuler(context):
```
tfoote, line 73: Please move this down to be in alphabetical order.

tfoote (changes requested):
I can do the proper review now with line comments.

wuxiaoqiang12:
Hi @Z572, @tfoote and the @ros-infrastructure maintenance team,

First, thank you for the feedback and the detailed review on this PR.

> Since there has been no activity here for a while, I have taken over this effort to ensure the openEuler support gets merged. I've addressed the feedback regarding the alphabetical ordering of registration and definitions in a new PR: #1021.

The new PR supersedes this one. Thank you @Z572 for the great initial work to get this started!

PR #920, opened 2023-05-05
add python3 bookworm targets

tfoote:
This will need a release or manual flood to deploy


cottsay:
This was done as part of #938

PR #921, opened 2023-05-10
Update rosdep help and logging for ROS 2.

cdhabecker:
[Problem]
- Git staging is cluttered with build files in install/ and log/.
- Argument <stacks-and-packages> (which relies on stack.xml and manifest.xml) does not make sense for ROS 2.
- Help text for --ignore-src is unclear whether it ignores the input pkgs or the dependency pkgs or both.
- The --verbose output does not show all the packages that rosdep thinks are installed (or available in the workspace).
- The --verbose output prints some lines out of order.
- The --verbose output prints dupes of keys, e.g., "resolve_all: rsource [pkg3] requires rosdep keys [pkg1 pkg1 pkg1]", followed by 3 identical lines describing the action.

[Solution]
- Modified .gitignore. Verified that the repo does not currently have any files in install/ or log/.
- Modified command usage. Split the check, install, and key commands into variants "ROS 1 or 2" and "ROS 1".
- Modified help text for --ignore-src. Added "--ignore-src:" marker to verbose log lines.
- Added verbose logging for --ignore-src that displays the packages that rosdep finds at AMENT_PREFIX_PATH_ENV_VAR. (The first part of --ignore-src, where it finds ROS_PACKAGE_PATH packages, already has verbose logging via find_catkin_packages_in().)
- Updated verbose logging in find_catkin_packages_in() so that it is on stdout and, therefore, prints in order with the other stdout lines.
- Updated resolve_all() to make the list of rosdep_keys unique.

[Test]
- Manual, local test on jammy-amd64.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@ dist
 src/rosdep.egg-info
 nose*
 _build
+install
+log
--- a/src/rosdep2/catkin_packages.py
+++ b/src/rosdep2/catkin_packages.py
@@ -25,12 +25,12 @@ def find_catkin_packages_in(path, verbose=False):
     if not os.path.exists(path):
         raise OSError("given path '{0}' does not exist".format(path))
     if verbose:
-        print("Looking for packages in '{0}'... ".format(path),
-              end='', file=sys.stderr)
+        print("Looking for catkin packages in '{0}'... ".format(path),
+              end='')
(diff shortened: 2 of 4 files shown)


cdhabecker:
See test output at https://gist.github.com/cdhabecker/250790b665240d2be8a24b1afe01eaae.

PR #923, opened 2023-06-26
Does not check python packages installed via GitHub url

canikrichard:
Does not check if the installation was successful when the python package was provided via URL, for example 
```
pip install <GITHUB_URL>
```
The URL does not contain the package name it is just a path to the setup.py.


cottsay:
Thanks makes changes the version parsing for every installer to accommodate an unsupported mechanism specific to pip.

I do not think we could make this change as-is.

PR #926, opened 2023-08-07
fix: pkg_resources deprecated warning

SubaruArai:
This pr fixes a warning caused by `pkg_resources`, by using `importlib.metadata` instead.

`Importlib.metadata` is supported since py3.8, which is shipped with the oldest supported ROS1/2 distro (noetic, Ubuntu 20.04)

[Same warning in colcon](https://github.com/colcon/colcon-core/issues/552) (maybe resolved)

### DIFF
--- a/src/rosdep2/platforms/pip.py
+++ b/src/rosdep2/platforms/pip.py
@@ -29,10 +29,14 @@
 
 from __future__ import print_function
 
-import importlib.metadata
 import os
 import subprocess
 import sys
+
+try:
+    import importlib.metadata as importlib_metadata
+except ImportError:
+    import importlib_metadata
 
 from ..core import InstallFailed
 from ..installers import PackageManagerInstaller
@@ -130,8 +134,8 @@ class PipInstaller(PackageManagerInstaller):
         super(PipInstaller, self).__init__(pip_detect, supports_depends=True)
 
     def get_version_strings(self):
-        pip_version = importlib.metadata.version('pip')
-        setuptools_version = importlib.metadata.version('setuptools')
+        pip_version = importlib_metadata.version('pip')
+        setuptools_version = importlib_metadata.version('setuptools')
         version_strings = [
             'pip {}'.format(pip_version),
             'setuptools {}'.format(setuptools_version),
--- a/src/rosdep2/platforms/pip.py
+++ b/src/rosdep2/platforms/pip.py
@@ -135,6 +135,7 @@ class PipInstaller(PackageManagerInstaller):
 
     def get_version_strings(self):
         pip_version = importlib_metadata.version('pip')
+        # keeping the name "setuptools" for backward compatibility
         setuptools_version = importlib_metadata.version('setuptools')
         version_strings = [
             'pip {}'.format(pip_version),
--- a/src/rosdep2/platforms/pip.py
+++ b/src/rosdep2/platforms/pip.py
@@ -29,8 +29,8 @@
 
 from __future__ import print_function
 
+import importlib.metadata
 import os
-import pkg_resources
 import subprocess
 import sys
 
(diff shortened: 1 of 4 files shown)


src/rosdep2/platforms/pip.py
```diff
@@ -29,8 +29,8 @@
 
 from __future__ import print_function
 
+import importlib.metadata
```
cottsay, line 32: The `importlib.metadata` module is only present in Python 3.8 and newer. This package must support Python 3 versions as far back as 3.6 (for RHEL 8).

The suggested pattern here is used in various other packages to prefer the Python 3.8+ module if present, and fall back to the backport package if it can't be found.

```suggestion
try:
    from importlib.metadata import version
except ImportError:
    from importlib_metadata import version
```

You'll also need to add this to the `install_requires`:   `importlib-metadata; python_version < "3.8"`

...and update the implementation to simply call `version` instead of `importlib.metadata.version`.

cottsay (changes requested):

SubaruArai:
@cottsay I've added importlib_metadata for py3.6.

Couple of questions:
1. should I change [stdeb.cfg](https://github.com/ros-infrastructure/rosdep/blob/master/stdeb.cfg) and bump the minimum OS and python version? Or should I add importlib_metadata to it?
2. should we drop setuptools from the dependency? AFAIK, it was added for `pkg_resources`. #809

SubaruArai:
@cottsay ping

cottsay:
Good questions.

> 1. should I change [stdeb.cfg](https://github.com/ros-infrastructure/rosdep/blob/master/stdeb.cfg) and bump the minimum OS and python version? Or should I add importlib_metadata to it?

Should be safe to add `python3 (>= 3.8) | python3-importlib-metadata` as a dependency to the `stdeb.cfg` [like this](https://github.com/colcon/colcon-core/blob/efc8d8d9ed531deb5bd549e8ff297974b92754e4/stdeb.cfg#L3C78-L3C123).

> 2. should we drop setuptools from the dependency? AFAIK, it was added for `pkg_resources`. [Require setuptools, src/rosdep2/platforms/pip.py imports pkg_resources #809](https://github.com/ros-infrastructure/rosdep/pull/809)

Yes, please do. Both `setup.cfg` and `stdeb.cfg` reference it.

SubaruArai:
@cottsay Couple of comments / questions (they'll be in their own pr if they need to be addressed)

1. I left `get_version_strings`'s output as `setuptools x.y.z`, for backwards compatibility. Or should I change it?
2. [stdeb.cfg](https://github.com/ros-infrastructure/rosdep/blob/master/stdeb.cfg) contains python2 related config. Should we delete it like in [colcon-core](https://github.com/colcon/colcon-core/blob/efc8d8d9ed531deb5bd549e8ff297974b92754e4/stdeb.cfg#L2C1-L2C12)?
3. [the ci file](https://github.com/ros-infrastructure/rosdep/blob/master/.github/workflows/ci.yaml) contains python2 tests. Should we delete it?

cottsay:
Thanks for iterating.
 
> 1. I left `get_version_strings`'s output as `setuptools x.y.z`, for backwards compatibility. Or should I change it?

I think it's good as-is. The setuptools version is useful here not because we're using setuptools (pkg_resources) to get the information, but because at the time, the packages used setuptools to build. While there are more options now and pip supports them, setuptools is still widely used and could be useful information debugging things. 

> 2. [stdeb.cfg](https://github.com/ros-infrastructure/rosdep/blob/master/stdeb.cfg) contains python2 related config. Should we delete it like in [colcon-core](https://github.com/colcon/colcon-core/blob/efc8d8d9ed531deb5bd549e8ff297974b92754e4/stdeb.cfg#L2C1-L2C12)?
> 3. [the ci file](https://github.com/ros-infrastructure/rosdep/blob/master/.github/workflows/ci.yaml) contains python2 tests. Should we delete it?

These need to be separate changes for sure, and I think they should be made prior to merging this change so that we can point to the exact moment we dropped Python 2 support from rosdep. I've [started this process](https://github.com/ros-infrastructure/catkin_pkg/pull/354) in `catkin_pkg`, and there will likely be a bit more to it than changing `ci.yaml` and `stdeb.cfg`. Can you give me a few days to work through that process?

SubaruArai:
Ping me once it's merged, I'll fix merge conflicts. (if any)

cottsay:
Can we add a simple test to exercise this code? Something like this could be added to `test_rosdep_pip.py`:
```python
def test_PipInstaller_get_version_strings():
    from rosdep2.platforms.pip import PipInstaller
    installer = PipInstaller()
    assert installer.get_version_strings()
```

SubaruArai:
@cottsay Done!

cottsay (approved):
Thanks for iterating

PR #928, opened 2023-08-17
fix flake8 errors

SubaruArai:

PR #929, opened 2023-08-17
fix ci: test_rosdep_catkin_support.py errors

SubaruArai:
- Changed old test target (ubuntu lucid: 10.04) to ubuntu focal: 20.04
- Resolving target for python bumped to python3

Related: #928 
With both merged, the ci should be fixed

### DIFF
--- a/test/test_rosdep_catkin_support.py
+++ b/test/test_rosdep_catkin_support.py
@@ -8,11 +8,11 @@ import pytest
 def test_workflow():
     try:
         installer = get_installer(APT_INSTALLER)
-        view = get_catkin_view('fuerte', 'ubuntu', 'lucid')
-        resolved = resolve_for_os('cmake', view, installer, 'ubuntu', 'lucid')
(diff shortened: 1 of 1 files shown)


peci1 (approved):
LGTM, thanks.

This will pop up again a few months/years after ROS 1 EOL. Is there a better way to check this, e.g. using rolling?

SubaruArai:
I have no idea.
It seemed to me that this CI error was introduced by removing the `lucid` target from the test database. (I forgot which pr) But then again, I didn't fully look into the under-the-hood behavior.

I think in an ideal world, CI testing data should be frozen unless the schema has changed - although I don't have any insight in how to do so (or if it is already!).

cottsay:
I'm not really sure of the details, but this doesn't appear to be necessary to make CI pass. How did you produce the error this change intends to solve?

SubaruArai:
@cottsay If the CI is passing now, some change else must have solved this issue instead.
It's been a long time so I can't say what, but I think we can close this now.

Issue #912, opened 2023-02-14
flake8 line length linting error regression


tfoote:
I just submited a PR #911 with a small change, and I'm getting tons of linting errors on Macos-latest python 3.10 due to line length.

It appears that the default has changed or the parameters to flake8 have changed so much of the code base is failing at 80 char width.

https://github.com/ros-infrastructure/rosdep/actions/runs/4178200482/jobs/7236711419
<details> 

```
>       assert not report.total_errors, \
            'flake8 reported {report.total_errors} errors'.format(**locals())
E       AssertionError: flake8 reported 714 errors
E       assert not 714
E        +  where 714 = <flake8.api.legacy.Report object at 0x1128bdf00>.total_errors

test/test_flake8.py:62: AssertionError
----------------------------- Captured stdout call -----------------------------

714   E501 line too long (86 > 79 characters)
----------------------------- Captured stderr call -----------------------------
/Users/runner/work/rosdep/rosdep/doc/rosdep_doc_utils.py:35:80: E501 line too long (86 > 79 characters)
                [py, '-c', "from rosdep2.main import rosdep_main;rosdep_main(['-h'])"]
                                                                               ^
/Users/runner/work/rosdep/rosdep/doc/rosdep_doc_utils.py:38:80: E501 line too long (93 > 79 characters)
                nodes.literal_block(text=re.sub(escaped_capitalized_usage, '', out.decode()))
                                                                               ^
/Users/runner/work/rosdep/rosdep/doc/rosdep_doc_utils.py:42:80: E501 line too long (97 > 79 characters)
                [py, '-c', "from rosdep2.main import rosdep_main;rosdep_main(['install', '-h'])"]
                                                                               ^
/Users/runner/work/rosdep/rosdep/doc/rosdep_doc_utils.py:45:80: E501 line too long (93 > 79 characters)
                nodes.literal_block(text=re.sub(escaped_capitalized_usage, '', out.decode()))
                                                                               ^
/Users/runner/work/rosdep/rosdep/setup.py:53:80: E501 line too long (98 > 79 characters)
    kwargs['install_requires'] += ['catkin_pkg >= 0.4.0', 'rospkg >= 1.4.0', 'rosdistro >= 0.7.5']
                                                                               ^
/Users/runner/work/rosdep/rosdep/src/rosdep2/__init__.py:49:80: E501 line too long (84 > 79 characters)
# don't let import error take down code as when attempting to compute version number
                                                                               ^
[...]

PR #930, opened 2023-08-17
Use setup.cfg to configure flake8, instead of in the ci code

SubaruArai:
This makes flake8's configuration more discoverable and easier to use with other tools (IDE, etc.).

~~Regression: it seems that "max-complexity" wasn't properly enforced, now a lot of code is detected as too complex. (warn if > 10, max is 25)
We may need to adjust the value.~~
Complexity adjusted to 26 to satisfy the current code's needs.

### DIFF
--- a/test/test_flake8.py
+++ b/test/test_flake8.py
@@ -14,7 +14,7 @@
 
 from __future__ import print_function
 
-import pathlib
+import os
 import subprocess
 import sys
 
@@ -27,14 +27,10 @@ def test_flake8():
     # flake8 doesn't have a stable public API as of ver 6.1.0.
     # See: https://flake8.pycqa.org/en/latest/user/python-api.html
     # Calling through subprocess is the most stable way to run it.
-    result = subprocess.run(
-        [sys.executable, '-m', 'flake8'],
-        stdout=subprocess.PIPE,  # capture_output doesn't work on py3.6
-        stderr=subprocess.PIPE,
+
+    # We still need to support Python 2.7, so we can't use run()
+    ret_code = subprocess.call(
+        [sys.executable, "-m", "flake8"],
         cwd=os.path.dirname(os.path.dirname(__file__)),
-        check=False,
     )
-    # pipe stdout to stdout, stderr to stderr
-    print(result.stdout.decode('utf-8'), end='')
-    print(result.stderr.decode('utf-8'), end='', file=sys.stderr)
-    assert 0 == result.returncode, "flake8 found violations"
+    assert 0 == ret_code, "flake8 found violations"
--- a/test/test_flake8.py
+++ b/test/test_flake8.py
@@ -28,10 +28,10 @@ def test_flake8():
     # See: https://flake8.pycqa.org/en/latest/user/python-api.html
     # Calling through subprocess is the most stable way to run it.
     result = subprocess.run(
-        ['flake8'],
+        [sys.executable, '-m', 'flake8'],
         stdout=subprocess.PIPE,  # capture_output doesn't work on py3.6
         stderr=subprocess.PIPE,
-        cwd=pathlib.Path(__file__).parent.parent,
+        cwd=os.path.dirname(os.path.dirname(__file__)),
         check=False,
     )
     # pipe stdout to stdout, stderr to stderr
--- a/test/test_flake8.py
+++ b/test/test_flake8.py
@@ -29,8 +29,9 @@ def test_flake8():
     # Calling through subprocess is the most stable way to run it.
     result = subprocess.run(
         ['flake8'],
+        stdout=subprocess.PIPE,  # capture_output doesn't work on py3.6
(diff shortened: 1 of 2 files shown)


SubaruArai:
@cottsay What do you say? There are dozens of functions flagged as "too complex"(C901).

I suggest changing `max-complexity` to 25 or 30.

SubaruArai:
We might as well revert #912 if this fixes those problems on macos. It might be related to complexity not being properly checked.

SubaruArai:
Just FYI, bumping `flake8` to `6.1.0` added these errors:

- E721: do not compare types, for exact checks use `is` / `is not`, for instance checks use `isinstance()`
- E231: missing whitespace after ','

Personally, I think both are valid and should be fixed in a separate PR.

cottsay:
If we're not enforcing the rule right now, the first move is to set a backstop at a higher level and get it merged before things get worse.

We can clean up the complexity and reduce the limit in a future change.

SubaruArai:
@cottsay by backstop, do you mean changing the complexity in this pr from 10 to around (whatever smallest that works)?

cottsay:
> do you mean changing the complexity in this pr from 10 to around (whatever smallest that works)?

Yes

test/test_flake8.py
```diff
@@ -14,50 +14,27 @@
...
-            'W504',  # ignore line breaks after binary operator (new rule added in 2018)
-        ],
-        max_line_length=200,
-        max_complexity=10,
-        show_source=True,
+    # flake8 doesn't have a stable public API as of ver 6.1.0.
+    # See: https://flake8.pycqa.org/en/latest/user/python-api.html
+    # Calling through subprocess is the most stable way to run it.
+    result = subprocess.run(
+        ['flake8'],
```
cottsay, line 31: The `flake8` executable may use a different interpreter than the one used to invoke these tests (and the one for which dependencies were installed for).
```suggestion
        [sys.executable, '-m', 'flake8'],
```

test/test_flake8.py
```diff
@@ -14,50 +14,27 @@
...
+        stdout=subprocess.PIPE,  # capture_output doesn't work on py3.6
+        stderr=subprocess.PIPE,
+        cwd=pathlib.Path(__file__).parent.parent,
```
cottsay, line 34: We haven't quite 100% abandoned Python 2 in `rosdep` yet (I know we don't have CI for it anymore).

```suggestion
        cwd=os.path.dirname(os.path.dirname(__file__)),
```

test/test_flake8.py
```diff
@@ -14,50 +14,27 @@
...
-        max_line_length=200,
-        max_complexity=10,
-        show_source=True,
+    # flake8 doesn't have a stable public API as of ver 6.1.0.
+    # See: https://flake8.pycqa.org/en/latest/user/python-api.html
+    # Calling through subprocess is the most stable way to run it.
+    result = subprocess.run(
+        ['flake8'],
+        stdout=subprocess.PIPE,  # capture_output doesn't work on py3.6
+        stderr=subprocess.PIPE,
```
cottsay, line 33: Is there any reason to capture the output? Maybe we should just let it bleed through to the parent process.

```suggestion
        stdout=subprocess.PIPE,  # capture_output doesn't work on py3.6
        stderr=subprocess.PIPE,
```

cottsay: Additionally, `subprocess.run` isn't available in Python 2, so we can use `subprocess.call` instead, which returns the process exit code instead of a CompletedProcess (since we're not capturing the output anymore, we don't really care).

cottsay (changes requested):
Thanks for tackling this. Only major concern is that we'd like to continue to support Python 2 until the next release.

Out of curiosity, why wasn't `max_complexity` enforced before? Underscore vs hyphen?

SubaruArai:
> Out of curiosity, why wasn't max_complexity enforced before? Underscore vs hyphen?

Looks like the [legacy API](https://flake8.pycqa.org/en/2.6.0/api.html) accepted `complexity` when running `check_file()`. Maybe this was the reason?

test/test_flake8.py
SubaruArai: > Is there any reason to capture the output? Maybe we should just let it bleed through ...

It was a long while ago, so I forgot why. I tested and there wasn't any difference, I think we can delete this capture.

> Python 2 support

Will do.

nuclearsandwich:
> That said, I'd like to hear from other `ros-infrastructure` maintainers if they see any drawbacks with the process invocation approach taken here.

I tested locally with both passing and non-passing result and in both cases the flake8 output was properly set.

This might even make it easier for me to add some local overrides (i.e. I have a `.direnv` folder in the repository that also needs to be excluded but haven't just caved and made it part of the file's default).

nuclearsandwich (approved):

PR #942, opened 2024-02-16
Bump to the latest GitHub Actions releases

cottsay:
- actions/checkout v2 -> v4
- actions/setup-python v1 -> v4


nuclearsandwich (approved):
It's probably worth a backlog to look at a renovatebot configuration which could keep us on this treadmill more easily.

PR #949, opened 2024-04-10
Implement test fixture for faking the rosdistro repo

cottsay:
This test was previously silently skipped when there wasn't an initialized and updated rosdep database present on the host. This change creates a 'fake' rosdep database locally and uses that to run the test. This decoupling also means that the test can run offline.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -1,0 +1,93 @@
+# Copyright 2024 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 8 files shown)


cottsay:
Alternatively, if we want to keep this test as a sort of "outer loop integration test", I can update these changes to use the live rosdep database but redirect only the sources list and cache to a temporary directory to isolate the test from the host.

If we do that, we'll need to continue to ensure that the live rosdep data is compatible with the test.

tfoote (approved):
Testing based on the mocked data makes more sense and is much better to have to make it repeatable. 

Maybe a method or comment how to update the content. Aka `# This content is mocking https://github.com/ros/rosdistro/rosindex-v4.yaml periodically this should be updated to a more recent snapshot`.

PR #950, opened 2024-04-10
Re-enable codecov upload

cottsay:
Fixes #947


christophebedard:
There doesn't seem to be anything here: https://app.codecov.io/gh/ros-infrastructure/rosdep/pull/950

nuclearsandwich (approved):
I'm good with merging this to get it on the default branch, which might kick on codecov for PRs. If nothing comes of this we can figure out if we need to authorize codecov at the org level or something.

PR #952, opened 2024-04-15
Fix potential bug in test_rosdep_sources_list.py

SubaruArai:
When running pytest from anywhere else than the project root directory, test_rosdep_sources_list.py would fail because there were 2 hardcoded path that depended on where the test was run from.

This commit fixes this by using a relative path from the aforementioned file, which is the same logic used in other tests.

Related: #951


cottsay (approved):
Awesome, thanks!

Issue #940, opened 2024-02-09
Apt install one-shot


tonynajjar:
This is more of a suggestion than an issue. I noticed that rosdep runs each apt install individually for each package which is probably more time consuming than running 'apt install package1 package2....'. Can we install them all in one-shot? This small change could save a lot of CI time for a lot of people.

What do you think?

SubaruArai:
@tonynajjar Made a PR, care to test it out?

tonynajjar:
hi @SubaruArai, thanks for tackling this! I tested it out and for my use case it accelerated the rosdep installation by 30 seconds which is a big improvement!

I'm just curious about one thing I see in the logs:

```
#28 27.51 executing command [sudo -H apt-get install -y python3-imageio python3-mock python3-pytest-cov ros-humble-foxglove-bridge ros-humble-foxglove-msgs ros-humble-nav-2d-msgs ros-humble-nav-2d-utils ros-humble-nav2-behavior-tree ros-humble-nav2-core ros-humble-nav2-costmap-2d ros-humble-phidgets-spatial ros-humble-realsense2-camera-msgs ros-humble-robot-localization ros-humble-rtcm-msgs ros-humble-rviz-imu-plugin ros-humble-rviz-satellite]
#28 27.51 executing command [sudo -H pip3 install -U utm]
#28 27.51 executing command [sudo -H apt-get install -y espeak iputils-ping python3-pygame python3-pykdl python3-shapely python3-sqlalchemy ros-humble-plotjuggler-ros ros-humble-rosbag2-storage-mcap]
#28 27.51 #All required rosdeps installed successfully
#28 DONE 27.7s
```

I understand why `pip3 install` done seperately of course but I wonder why there are still 2 `apt-get` commands instead of one. It's already a huge improvement from:

```
#38 58.68 executing command [sudo -H apt-get install -y iputils-ping]
#38 58.68 executing command [sudo -H apt-get install -y ros-humble-foxglove-bridge]
#38 58.68 executing command [sudo -H apt-get install -y ros-humble-realsense2-camera-msgs]
#38 58.68 executing command [sudo -H apt-get install -y python3-mock]
#38 58.68 executing command [sudo -H apt-get install -y ros-humble-rtcm-msgs]
#38 58.68 executing command [sudo -H apt-get install -y ros-humble-robot-localization]
#38 58.68 executing command [sudo -H apt-get install -y ros-humble-phidgets-spatial]
[...]

PR #953, opened 2024-04-15
Feature: Apt install command to install all packages at once

SubaruArai:
See: #940

This should make the installation faster.
If this is successful, we might need to think about other package managers as well.

### DIFF
--- a/src/rosdep2/platforms/debian.py
+++ b/src/rosdep2/platforms/debian.py
@@ -294,13 +294,11 @@ class AptInstaller(PackageManagerInstaller):
         return ['apt-get {}'.format(version)]
 
     def _get_install_commands_for_package(self, base_cmd, package_or_list):
-        def pkg_command(p):
-            return self.elevate_priv(base_cmd + [p])
-
-        if isinstance(package_or_list, list):
-            return [pkg_command(p) for p in package_or_list]
+        if isinstance(package_or_list, str):
+            return self.elevate_priv(base_cmd + [package_or_list])
         else:
-            return pkg_command(package_or_list)
+            # sort to make the output deterministic
+            return self.elevate_priv(base_cmd + sorted(package_or_list))
 
     def get_install_command(self, resolved, interactive=True, reinstall=False, quiet=False):
         packages = self.get_packages_to_install(resolved, reinstall=reinstall)
@@ -312,4 +310,9 @@ class AptInstaller(PackageManagerInstaller):
         if quiet:
             base_cmd.append('-qq')
 
-        return [self._get_install_commands_for_package(base_cmd, p) for p in _iterate_packages(packages, reinstall)]
+        return [
+            self._get_install_commands_for_package(
+                base_cmd,
+                _iterate_packages(packages, reinstall),
+            )
+        ]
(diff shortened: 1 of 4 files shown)


SubaruArai:
> The original reason that rosdep behaves in this way is to ensure that packages will still get installed on a best-effort basis despite an issue with one of them.
> ...
> At the time when apt support was implemented, there didn't appear to be a way to achieve that behavior.

Interesting, thanks for the input!
Although I'm not really sure if I like this decision, I can get it.
(I personally prefer all or nothing (atomic) - which isn't achievable since we're using multiple package managers)

Anyways, we should be able to add such a feature with [something like this](https://askubuntu.com/a/1333505).

I guess the real question will be - do we want to preserve this behavior, or speed-up the installation instead (especially in CI/CD systems)?
According to @tonynajjar, a 24 apt package + 1 pip package installation was reduced from 58s to 28s - which I think is a big deal in the context of CI systems. I also prefer failing fast (it looks like apt-get fails at the lookup stage) rather than figuring out one package went wrong after a while.

@cottsay What do you say? Keep the best-effort behavior, or ditch it for faster execution?

tonynajjar:
I would really like to keep both features if possible. In the package.xml I also define dependencies on custom packages that are not released, for the purpose of colcon building. So when running rosdep, it fails to fetch these cuatom packages but still continues with the rest.
As for faster rosdep, it's obvious why we would want that - imagine all the CI time you would save in total across the world 😬

tonynajjar:
@cottsay did you have the chance to think about it?

cottsay:
> Keep the best-effort behavior, or ditch it for faster execution?

I do not believe that changing the existing best-effort behavior is a good idea, and I can't see myself approving a PR that changes that only to improve performance.

A change to a single-invocation that preserves the established behavior and keeps `apt` aligned with the other common package managers supported by rosdep would certainly get my attention.

tonynajjar:
There is already the flag `-r Continue installing despite errors`, can't we apply the "one-shot" approach only if this flag is on?

SubaruArai:
It's been a while and I can't recall what I tried, but I remember that there wasn't a silver bullet - including the `--ignore-missing` flag.

130s (commented):
I reviewed the code. The current state of this PR would change `rosdep`'s behavior as already discussed in the review comments section. I +1 for not changing the default behavior, but also personally +1 for adding an optional feature to faster installation.

I'm not a maintainer.

aaron-riact:
this seems nice.  can it be taken out of draft now?

SubaruArai:
@aaron-riact Sry for the half-assed and abandoned pr.

1. At its current state, I cannot take it out of a draft, nor will @cottsay approve due to aforementioned behavior changes.
   (and I agree with him)
2. I think the middle ground will be:
2.1. Add a new option (e.g. `--fail-fast`) to `rosdep install`
2.2. Implement a faster install method, (while still half-assed, see https://github.com/ros-infrastructure/rosdep/issues/940#issuecomment-2056413128) to be triggered by above option
2.3. Implement relevant test codes

There's one thing that I haven't decided though - what should the option's alias be?
I don't like using `-f`, my instinct would be `--force` for `-f`.
A somewhat better (though still ugly imo) proposal would be: `-F --fail-fast`

What do you all say?

krisl:
> There's one thing that I haven't decided though - what should the option's alias be?

Given this seems most useful for shaving off those precious CI seconds, probably we can afford an awkward long option. How about `--single-install-command=apt` or similar, with no short alias?

krisl:
@SubaruArai cool.  Will you make the changes needed to add the optional behaviour?
And can someone comment if that change would be sufficient to get this merged?

Issue #922, opened 2023-06-23
Installation instructions invalid for Ubuntu 23.04


Ryanf55:
Hello,

I am trying to use rosdep on Ubuntu 23.04 (and potentially contribute to making it work). 

The installation tutorials recommend installing with `sudo pip3`, but Ubuntu 23 strongly discourages. 
```
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.
    
    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.
    
    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.
    
    See /usr/share/doc/python3.11/README.venv for more information.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.

```
That said, the debian `python3-rosdep2` works fine. 

Is there a reason the docs don't recommend using a package manager like `apt` to manage dependencies on `apt`? In the ROS 2 tutorials, [they recommend](http://docs.ros.org/en/iron/Installation/Alternatives/Ubuntu-Development-Setup.html#install-development-tools) installing `ros-dev-tools.

What situations are the two methods (apt and pip) preferred? I'm happy to issue a PR once I get a recommendation. 

Related: https://github.com/ros2/ros2_documentation/issues/3757
Rosdep version: `0.22.2`

tfoote:
Fixed in #979

PR #954, opened 2024-04-15
Doc: recommend system install of rosdep

SubaruArai:
The install instructions for rosdep were recommending a pip install, which is not the recommended way to install rosdep. (#922)
This change updates the install instructions to recommend a system install of rosdep.

### DIFF
--- a/doc/overview.rst
+++ b/doc/overview.rst
@@ -4,6 +4,19 @@ Overview
 Installing rosdep
 -----------------
 
+.. admonition:: Note
+
+    If you want to use rosdep with ROS1/2, you should install rosdep
+    following their installation instructions:
+
+    * `ROS1 installation instructions
+      <http://wiki.ros.org/ROS/Installation>`_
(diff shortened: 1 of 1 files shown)


doc/overview.rst
```diff
@@ -4,12 +4,36 @@ Overview
...
-or::
+rosdep2 is a system package under Ubuntu and Debian::
 
+    # Ubuntu >= 20.04 (Focal)
+    sudo apt-get install python3-rosdep
+    # Debian >=11 (Bullseye)
+    sudo apt-get install python3-rosdep2
+
+If rosdep doesn't exist in your package manager, you can install it
+using pip or easy_install::
```
nuclearsandwich, line 30: It's probably also worth dropping the reference to easy_install as it is no longer tested or used in our CI stack to the best of my knowledge.

nuclearsandwich:
Merging as-is in order to take the improvement. Thanks for the contribution!

PR #973, opened 2024-06-20
Remove more Python 2 artifacts

cottsay:
We no longer support Python 2 in rosdep so there's no reason to carry these snippets anymore.


christophebedard (approved):
Looks good to me.

tfoote (approved):

PR #974, opened 2024-06-20
Updates to run and pass flake8-builtins

cottsay:
There are flake8-builtins violations which appear as part of rosdep's public API. We'll supporess those to avoid breaking API, but still add the linter plugin so we don't regress more.


christophebedard (approved):
Looks reasonable to me.

Issue #978, opened 2024-07-19
Rosdep fails to install pip packages on Python3.11+


InvincibleRMC:
In Python3.11+ rosdep fails to install pip packages because python wants virtual environments to be used. ROS does not yet support virtual environments yet.

```
  executing command [pip3 install -U -q pytest-qt]
  WARNING: Skipping /usr/lib/python3.12/dist-packages/argcomplete-3.1.4.dist-info due to invalid metadata entry 'name'
  WARNING: Skipping /usr/lib/python3.12/dist-packages/argcomplete-3.1.4.dist-info due to invalid metadata entry 'name'
  error: externally-managed-environment
  
  × This environment is externally managed
  ╰─> To install Python packages system-wide, try apt install
      python3-xyz, where xyz is the package you are trying to
      install.
      
      If you wish to install a non-Debian-packaged Python package,
      create a virtual environment using python3 -m venv path/to/venv.
      Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
      sure you have python3-full installed.
      
      If you wish to install a non-Debian packaged Python application,
      it may be easiest to use pipx install xyz, which will manage a
      virtual environment for you. Make sure you have pipx installed.
      
      See /usr/share/doc/python3.12/README.venv for more information.
  
  note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
  hint: See PEP 668 for the detailed specification.
  ERROR: the following rosdeps failed to install
    pip: command [pip3 install -U -q pytest-qt] failed
```

ros-discourse:
This issue has been mentioned on **ROS Discourse**. There might be relevant details there:

https://discourse.ros.org/t/rosdep-for-pip-is-broken-on-jazzy/38981/1

InvincibleRMC:
Closed by #979

PR #979, opened 2024-08-08
Check for pip configuration when an externally managed environment is detected.

nuclearsandwich:
rosdep is designed to treat pip like an alternative system-level package manager. Deviating from this approach is not easily achievable without a significant rethinking of how pip packages are managed. In the meantime, we can at least instruct users how to restore the prior functionality.

Rather than inject the environment variable / config on behalf of the user, this change instructs them to make the necessary config changes themself, keeping them informed of the change their making to the system's new default.

I added a documentation page specific to this issue with additional configuration details and recipes for CI.

Seeks to reduce friction related to #978

### DIFF
--- a/doc/pip_and_pep_668.rst
+++ b/doc/pip_and_pep_668.rst
@@ -3,7 +3,7 @@ Pip installation after PEP 668
 
 `PEP-668`_ introduced `externally managed environments <externally-managed-environments>`_ to Python packaging.
 
-Rosdep is designed to use pip as an alternative system package manager, rosdep installation of pip packages requires installing packages globally as root.
+``rosdep`` is designed to use pip as an alternative system package manager, rosdep installation of pip packages requires installing packages globally as root.
 Starting with Python 3.11, `PEP-668`_ compliance requires you to allow pip to install alongside externally managed packages using the ``break-system-packages`` option.
 
 There are multiple ways to configure pip so that rosdep will succeed.
--- a/doc/pip_and_pep_668.rst
+++ b/doc/pip_and_pep_668.rst
@@ -27,7 +27,7 @@ If your system's sudo configuration prohibits the passing of environment variabl
 Configure using pip.conf
 ------------------------
 
-`Pip configuration files <pip-configuration>`_ can be used to set the desired behavior 
+`Pip configuration files <pip-configuration>`_ can be used to set the desired behavior.
 Pip checks for global configuration files in ``XDG_CONFIG_DIRS``, as well as ``/etc/pip.conf``
 For details on ``XDG_CONFIG_DIRS`` refer to the `XDG base directories specification <xdg-base-dirs>`_.
 If you're unsure which configuration file is in use by your system, ``/etc/pip.conf`` seems like the most generic.
--- a/doc/pip_and_pep_668.rst
+++ b/doc/pip_and_pep_668.rst
@@ -28,13 +28,15 @@ Configure using pip.conf
 `Pip configuration files <pip-configuration>`_ can be used to set the desired behavior 
 Pip checks for global configuration files in ``XDG_CONFIG_DIRS``, as well as ``/etc/pip.conf``
 For details on ``XDG_CONFIG_DIRS`` refer to the `XDG base directories specification <xdg-base-dirs>`_.
-
-Creating a pip.conf in your user account's ``XDG_CONFIG_HOME`` (e.g. ``~/.config/pip/pip.conf``) does not appear to be sufficent when installing packages globally.
+If you're unsure which configuration file is in use by your system, ``/etc/pip.conf`` seems like the most generic.
 
 .. code-block:: ini
 
    [install]
    break-system-packages = true
+
+
+.. warning:: Creating a pip.conf in your user account's ``XDG_CONFIG_HOME`` (e.g. ``~/.config/pip/pip.conf``) does not appear to be sufficent when installing packages globally.
 
 
 Configuring for CI setup
--- a/doc/pip_and_pep_668.rst
+++ b/doc/pip_and_pep_668.rst
@@ -19,8 +19,10 @@ The value of the environment variable can be any of ``1``, ``yes``, or ``true``.
 The string values are not case sensitive.
 
 rosdep is designed to use ``sudo`` in order to gain root privileges for installation when not run as root.
-If your system's sudo configuration prohibits the passing of environment variables
+If your system's sudo configuration prohibits the passing of environment variables use the :ref:`pip.conf <configure-using-pip.conf>` method below.
 
+
+.. _configure-using-pip.conf:
 
 Configure using pip.conf
 ------------------------
--- a/doc/pip_and_pep_668.rst
+++ b/doc/pip_and_pep_668.rst
@@ -1,0 +1,67 @@
+Pip installation after PEP 668
+==============================
+
+`PEP-668`_ introduced `externally managed environments <externally-managed-environments>`_ to Python packaging.
+
+Rosdep is designed to use pip as an alternative system package manager, rosdep installation of pip packages requires installing packages globally as root.
+Starting with Python 3.11, `PEP-668`_ compliance requires you to allow pip to install alongside externally managed packages using the ``break-system-packages`` option.
+
+There are multiple ways to configure pip so that rosdep will succeed.
+
+
+Configure using environment variable
+------------------------------------
+
+This is the way that we recommend configuring pip for rosdep usage.
+We recommend configuring pip using the system environment.
+Setting environment variables in your login profile, ``PIP_BREAK_SYSTEM_PACKAGES`` in your environment.
+The value of the environment variable can be any of ``1``, ``yes``, or ``true``.
+The string values are not case sensitive.
+
+
+Configure using pip.conf
+------------------------
+
+`Pip configuration files <pip-configuration>`_ can be used to set the desired behavior 
+Pip checks for global configuration files in ``XDG_CONFIG_DIRS``, as well as ``/etc/pip.conf``
+For details on ``XDG_CONFIG_DIRS`` refer to the `XDG base directories specification <xdg-base-dirs>`_.
+
+Creating a pip.conf in your user account's ``XDG_CONFIG_HOME`` (e.g. ``~/.config/pip/pip.conf``) does not appear to be sufficent when installing packages globally.
+
+.. code-block:: ini
+
+   [install]
+   break-system-packages = true
+
+
+Configuring for CI setup
+------------------------
+
+Either environment variables or configuration files can be used with your CI system.
+Which one you choose will depend on how your CI environment is configured.
+Perhaps the most straightforward will be to set the environent variable in the shell or script execution context before invoking rosdep.
+
+.. code-block:: bash
+
+   sudo rosdep init
+   rosdep update
+   PIP_BREAK_SYSTEM_PACKAGES=1 rosdep install -r rolling --from-paths src/
+
+If rosdep is invoked by internal processes in your CI and you need to set the configuration without having direct control over how ``rosdep install`` is run, setting the environment variable globally would also work.
+
+.. code-block:: bash
+
+   export PIP_BREAK_SYSTEM_PACKAGES=1
+   ./path/to/ci-script.sh
+
+
+If you cannot set environment variables but you can create configuration files, you can set ``/etc/pip.conf`` with the necessary configuration.
+
+.. code-block:: bash
+
+   printf "[install]\nbreak-system-packages = true\n" | sudo tee -a /etc/pip.conf
+
+.. _PEP-668: https://peps.python.org/pep-0668/
+.. _pip-configuration: https://pip.pypa.io/en/stable/topics/configuration/
+.. _externally-managed-environments: https://packaging.python.org/en/latest/specifications/externally-managed-environments/
+.. _xdg-base-dirs: https://specifications.freedesktop.org/basedir-spec/latest/
--- a/doc/pip_and_pep_668.rst
+++ b/doc/pip_and_pep_668.rst
@@ -28,7 +28,7 @@ Configure using pip.conf
 ------------------------
 
 `Pip configuration files <pip-configuration>`_ can be used to set the desired behavior.
-Pip checks for global configuration files in ``XDG_CONFIG_DIRS``, as well as ``/etc/pip.conf``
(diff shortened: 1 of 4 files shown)


nuclearsandwich:
In addition to the test failures, this actually doesn't work as intended because `sudo` is stripping the environment. I'm hesitant to recommend the pip global config file as a default because the default location is variable. Alternatively, we could override the sudo_command for the pip installer to pass that environment variable through.

ros-discourse:

https://discourse.ros.org/t/rosdep-for-pip-is-broken-on-jazzy/38981/5

src/rosdep2/platforms/pip.py
```diff
@@ -43,6 +46,15 @@
 # pip package manager key
 PIP_INSTALLER = 'pip'
 
+EXTERNALLY_MANAGED_EXPLAINER = """
+rosdep installation of pip packages requires installing packages packages globally as root
```
christophebedard, line 50:
```suggestion
rosdep installation of pip packages requires installing packages globally as root.
```

src/rosdep2/platforms/pip.py
```diff
@@ -70,6 +82,35 @@ def get_pip_command():
...
+    managed" envrionments such operating systems with included package
+    managers. If we're on Python 3.11 or greater, we need to check that pip
+    is configured to allow installing system-wide packages with the
+    flagrantly named "break system packages" config option or environment
+    variable.
+    """
+    if sys.version_info[0] == 3 and sys.version_info[1] >= 11:
+        if "PIP_BREAK_SYSTEM_PACKAGES" in os.environ and os.environ[
+            "PIP_BREAK_SYSTEM_PACKAGES"
+        ].lower() in ["yes", "1", "true"]:
```
christophebedard, line 97:
```suggestion
        ].lower() in ("yes", "1", "true"):
```

nuclearsandwich:
> In addition to the test failures, this actually doesn't work as intended because `sudo` is stripping the environment. I'm hesitant to recommend the pip global config file as a default because the default location is variable. Alternatively, we could override the sudo_command for the pip installer to pass that environment variable through.

As of https://github.com/ros-infrastructure/rosdep/pull/979/commits/2c76686afa24ff09f52642a4b72116c2051ae5e3 the installer is configured to pass `PIP_BREAK_SYSTEM_PACKAGES` through from the user environment. This works by default on each of my tested sudo environments (Archlinux, Ubuntu, Fedora) but a locked down sudo that disallows environment preservation or has an explicit allowlist will still create problems. My hope was to detect that the check passed but the install failed due to the externally managed target anyway and give a hint about sudo to suggest an alternative configuration method. But without changes, the pip installer never gets control back from `install_resolved()` so we'd have to add pip-specific hooks elsewhere in the code and I'm hesitant. I think I'm content to leave it here but with more robust documentation.

src/rosdep2/platforms/pip.py
```diff
@@ -43,6 +46,17 @@
...
 
+EXTERNALLY_MANAGED_EXPLAINER = """
+rosdep installation of pip packages requires installing packages globally as root.
+When using Python >= 3.11, PEP 668 compliance requires you to allow pip to install alongside
+externally managed packages using the 'break-system-packages' option.
+The recommeded way to set this option when using rosdep is to set the environment variable
+PIP_BREAK_SYSTEM_PACKAGES=1
+in your environment.
+
+For more information refer to http://docs.ros.org/en/independent/api/rosdep/html/pip_and_pep_668.html
```
nuclearsandwich, line 57: Note that this documentation won't actually be live until the PR is merged and docs rebuild the next day. This will only affect those using rosdep from mainline or this branch. I'll wait until the docs rebuild to make a new rosdep release with this change.

doc/pip_and_pep_668.rst
```diff
@@ -0,0 +1,72 @@
...
+------------------------------------
+
+This is the way that we recommend configuring pip for rosdep usage.
+We recommend configuring pip using the system environment.
+Setting environment variables in your login profile, ``PIP_BREAK_SYSTEM_PACKAGES`` in your environment.
+The value of the environment variable can be any of ``1``, ``yes``, or ``true``.
+The string values are not case sensitive.
+
+``rosdep`` is designed to use ``sudo`` in order to gain root privileges for installation when not run as root.
+If your system's sudo configuration prohibits the passing of environment variables
```
christophebedard, line 22: it looks like this sentence was cut off.

doc/pip_and_pep_668.rst
```diff
@@ -0,0 +1,72 @@
...
+
+
+Configure using pip.conf
+------------------------
+
+`Pip configuration files <pip-configuration>`_ can be used to set the desired behavior
```
christophebedard, line 28:
```suggestion
`Pip configuration files <pip-configuration>`_ can be used to set the desired behavior.
```

doc/pip_and_pep_668.rst
```diff
@@ -0,0 +1,72 @@
...
+
+``rosdep`` is designed to use ``sudo`` in order to gain root privileges for installation when not run as root.
+If your system's sudo configuration prohibits the passing of environment variables
+
+
+Configure using pip.conf
+------------------------
+
+`Pip configuration files <pip-configuration>`_ can be used to set the desired behavior 
+Pip checks for global configuration files in ``XDG_CONFIG_DIRS``, as well as ``/etc/pip.conf``
```
christophebedard, line 29:
```suggestion
Pip checks for global configuration files in ``XDG_CONFIG_DIRS``, as well as ``/etc/pip.conf``.
```

doc/pip_and_pep_668.rst
```diff
@@ -0,0 +1,72 @@
+Pip installation after PEP 668
+==============================
+
+`PEP-668`_ introduced `externally managed environments <externally-managed-environments>`_ to Python packaging.
+
+``rosdep`` is designed to use pip as an alternative system package manager, rosdep installation of pip packages requires installing packages globally as root.
```
christophebedard, line 6: this document is a bit inconsistent with rosdep and `rosdep`, but I'll let you decide if you want to change anything.

src/rosdep2/platforms/pip.py
```diff
@@ -70,6 +84,41 @@ def get_pip_command():
     return None
 
 
+def externally_managed_installable():
+    """
+    PEP 668 enacted in Python 3.11 blocks pip from working in "externally
+    managed" envrionments such operating systems with included package
```
christophebedard, line 90:
```suggestion
    managed" environments such as operating systems with included package
```

src/rosdep2/platforms/pip.py
```diff
@@ -70,6 +84,41 @@ def get_pip_command():
...
+    managers. If we're on Python 3.11 or greater, we need to check that pip
+    is configured to allow installing system-wide packages with the
+    flagrantly named "break system packages" config option or environment
+    variable.
+    """
+    if sys.version_info >= (3, 11):
+        if 'PIP_BREAK_SYSTEM_PACKAGES' in os.environ and os.environ[
+            'PIP_BREAK_SYSTEM_PACKAGES'
+        ].lower() in ('yes', '1', 'true'):
```
christophebedard, line 99: minor suggestion, but, if you want, you could improve readability a little bit by changing it to something like:

```suggestion
        if (
            'PIP_BREAK_SYSTEM_PACKAGES' in os.environ
            and os.environ['PIP_BREAK_SYSTEM_PACKAGES'].lower() in ('yes', '1', 'true')
        ):
```

not actually sure if exact change passes linters, though.

src/rosdep2/platforms/pip.py
```diff
@@ -70,6 +84,41 @@ def get_pip_command():
...
+def externally_managed_installable():
+    """
+    PEP 668 enacted in Python 3.11 blocks pip from working in "externally
+    managed" envrionments such operating systems with included package
+    managers. If we're on Python 3.11 or greater, we need to check that pip
+    is configured to allow installing system-wide packages with the
+    flagrantly named "break system packages" config option or environment
+    variable.
+    """
+    if sys.version_info >= (3, 11):
```
christophebedard, line 96: you could flip this to return early and remove one indent level from the rest of the function

```suggestion
    if sys.version_info < (3, 11):
        return True
```

test/test_rosdep_pip.py
```diff
@@ -67,6 +67,7 @@ def test_PipInstaller_get_depends():
     assert ['foo'] == installer.get_depends(dict(depends=['foo']))
 
 
+@patch.dict(os.environ, {'PIP_BREAK_SYSTEM_PACKAGES': '1'})
```
christophebedard, line 70: would be nice to test that the command fails if this isn't set and the version is >= 3.11, but you'd also need to make sure that the other config methods are not used, or just patch `externally_managed_installable()`, but I would leave that up to you.

christophebedard (commented):
Minor comments!

doc/pip_and_pep_668.rst
nuclearsandwich: My mental style guide is `rosdep` when referring to the CLI tool and rosdep when referring to the project. I am always tempted into using `rosdep` in appropriately when its the first word of the sentence because Rosdep and ROSdep are both wrong.

I will reword to avoid the entire problem :D (that mainly exists in my head)

nuclearsandwich: Actually, I'm just going to get over myself and start the sentence with the proper lowercase noun. It's the 2020s.

doc/pip_and_pep_668.rst
nuclearsandwich: lol good catch. I'm sure that I went to look up how to reference a section in the doc below and got distracted because that's exactly what I just did again.

src/rosdep2/platforms/pip.py
nuclearsandwich: The only think flake8 got upset by was the and being on the continuing line rather than the end of the first. I like this form better thanks.

test/test_rosdep_pip.py
nuclearsandwich: I'm limited somewhat by my lack of experience with Python's mocking. Let me see if I can skim enough examples and read enough of the docs to stub the method for a few tests before.

It's a good push. Tests added in 8b48f13

test/test_rosdep_pip.py
```diff
@@ -67,6 +67,23 @@ def test_PipInstaller_get_depends():
     assert ['foo'] == installer.get_depends(dict(depends=['foo']))
 
 
+@patch('rosdep2.platforms.pip.externally_managed_installable')
+def test_PipInstaller_handles_externally_managed_environment(externally_managed_installable):
+    from rosdep2 import InstallFailed
+    from rosdep2.platforms.pip import EXTERNALLY_MANAGED_EXPLAINER, PipInstaller
```
nuclearsandwich, line 73: I'm not super up on why imports are done per test function here but I am propagating the convention until I understand it better.

src/rosdep2/platforms/pip.py
nuclearsandwich: This suggestion was incorporated into a2e9a1e

christophebedard: I think you have an extra indentation level there:

```diff
     if (
-             'PIP_BREAK_SYSTEM_PACKAGES' in os.environ and
-             os.environ['PIP_BREAK_SYSTEM_PACKAGES'].lower() in ('yes', '1', 'true')
+         'PIP_BREAK_SYSTEM_PACKAGES' in os.environ and
+         os.environ['PIP_BREAK_SYSTEM_PACKAGES'].lower() in ('yes', '1', 'true')
     ):
         return True
```

in my experience the usual linters don't complain about this for some reason.

> The only think flake8 got upset by was the and being on the continuing line rather than the end of the first.

Ah yeah I always get bit by that!

test/test_rosdep_pip.py
```diff
@@ -67,6 +67,23 @@ def test_PipInstaller_get_depends():
...
+
+    externally_managed_installable.return_value = False
+    installer = PipInstaller()
+    try:
+        installer.get_install_command(['whatever'])
+        assert False, 'should have raised'
+    except InstallFailed as e:
+        assert e.failures == [('pip', EXTERNALLY_MANAGED_EXPLAINER)]
```
christophebedard, line 81: If you want, a more Pythonic way could be:

```suggestion
    with pytest.raises(InstallFailed) as exception_info:
        installer.get_install_command(['whatever'])
    assert exception_info.value.failures == [('pip', EXTERNALLY_MANAGED_EXPLAINER)]
```

and make sure to `import pytest`

christophebedard (approved):
Aside from my last minor suggestions, this looks good to me!

src/rosdep2/platforms/pip.py
```diff
@@ -70,6 +84,45 @@ def get_pip_command():
...
+    if 'XDG_CONFIG_DIRS' in os.environ:
+        for xdg_dir in os.environ['XDG_CONFIG_DIRS'].split(':'):
+            pip_config_file = Path(xdg_dir) / 'pip' / 'pip.conf'
+            pip_config.read(pip_config_file)
+            if pip_config['install']['break-system-packages']:
+                return True
+
+    fallback_config = Path('/etc/pip.conf')
+    pip_config.read(fallback_config)
+    if pip_config['install']['break-system-packages']:
```
christophebedard, line 119: I gave it a try and it seems like, if:

1. the config file doesn't exist; or
2. the `[install]` section doesn't exist; or
3. a `break-system-package` value isn't specified.

then `pip_config[...]` raises a `KeyError`. Looks like you could avoid checking all of that by using `getboolean()`:

```py
pip_config.getboolean('install', 'break-system-packages', fallback=False)
```

test/test_rosdep_pip.py
nuclearsandwich: Thanks for the suggestion. I'm actually going to keep this as is for now, because I don't believe the infrastructure packages ever explicitly require pytest. I think pytest is supposed to be compatible with stdlib unittest as well and utilizing that may be the preferred approach.

src/rosdep2/platforms/pip.py
nuclearsandwich: Good catch!

nuclearsandwich: Resolved in [8e94ef1](https://github.com/ros-infrastructure/rosdep/pull/979/commits/8e94ef1ffb01394b23cea9dcc12fbb6fc92f10d2)

christophebedard (approved):

PR #990, opened 2024-10-17
Apt install all at once

MrBlenny:


tfoote (changes requested):
In general we could consider making a change to install . There's already a smaller change in https://github.com/ros-infrastructure/rosdep/pull/953 which has a bunch of discussion on it.  I'd suggest following up over there.

MrBlenny:
Oops sorry, this was meant to go into our fork 😅
Apt install all at once with version eq support.

PR #997, opened 2025-03-26
Added search command to search for rosdep keys.

StefanFabian:
I've had to open the rosdep repo one time too many to look for a package key because I couldn't guess the correct one.
Hence, I've added a search command for rosdep which fuzzy searches the sources for a key or system package matching the search terms.
It allows one error per search term provided after `rosdep search`.
This could probably be made a parameter but typically this should be enough.

Examples:

```bash
$ rosdep search libpcl-dev
Closest keys:
  libfcl-dev

Closest packages:
  libpcl-all-dev: libpcl-dev [debian]
  libpcl-all-dev: libpcl-dev [ubuntu]
  acl: libacl-devel [fedora]
  acl: libacl-devel [rhel]

```

```bash
$ rosdep search pinoccio
Closest keys:
  kinematics_interface_pinocchio
  pinocchio

```

### DIFF
--- a/test/test_rosdep_main.py
+++ b/test/test_rosdep_main.py
@@ -286,14 +286,14 @@ class TestRosdepMain(unittest.TestCase):
 
         try:
             with fakeout() as b:
-                rosdep_main(['search', 'curl'] + cmd_extras)
+                rosdep_main(['search', 'curl', '--os=debian:squeeze'] + cmd_extras)
                 stdout, stderr = b
                 assert 'Closest keys' in stdout.getvalue(), stdout.getvalue()
                 assert 'curl' in stdout.getvalue(), stdout.getvalue()
                 assert 'Closest packages' not in stdout.getvalue(), stdout.getvalue()
                 assert not stderr.getvalue(), stderr.getvalue()
             with fakeout() as b:
-                rosdep_main(['search', 'libeigen3-dev'] + cmd_extras)
+                rosdep_main(['search', 'libeigen3-dev', '--os=ubuntu:noble'] + cmd_extras)
                 stdout, stderr = b
                 assert 'Closest keys' not in stdout.getvalue(), stdout.getvalue()
                 assert 'Closest packages' in stdout.getvalue(), stdout.getvalue()
@@ -301,6 +301,11 @@ class TestRosdepMain(unittest.TestCase):
                 assert not stderr.getvalue(), stderr.getvalue()
         except SystemExit:
             assert False, 'system exit occurred'
+        try:
+            rosdep_main(['search', 'libeigen3-dev', '--os=debian:squeeze'] + cmd_extras)
+            assert False, 'system exit should have occurred'
+        except SystemExit:
+            pass
         try:
             rosdep_main(['search', 'nonexistent'] + cmd_extras + ['-i'])
             assert False, 'system exit should have occurred'


ros-discourse:

https://discourse.ros.org/t/rosdep-search/43148/1

src/rosdep2/main.py
```diff
@@ -918,6 +922,80 @@ def command_resolve(args, options, lookup=None):
...
+            sorted_pkgs = list(filter(lambda x: x['error'] == 0, sorted_pkgs))
+            has_exact_match = True
+    if len(sorted_pkgs) > 0:
+        print('Closest packages:')
+        if not has_exact_match and len(sorted_pkgs) > 10:
+            sorted_pkgs = sorted_pkgs[:10]
+        for pkg in sorted_pkgs:
+            print('  %s: %s [%s]' % (pkg['key'], pkg['pkg'], pkg['os']))
+        if not has_exact_match and len(close_keys) > 10:
+            print('  [and %d more]' % (len(close_keys) - 10))
```
peci1, line 991:
```suggestion
        if not has_exact_match and len(close_pkgs) > 10:
            print('  [and %d more]' % (len(close_pkgs) - 10))
```

peci1: Or you could also use `sorted_pkgs` here to be a little more reader-friendly.

peci1 (approved):
I've tested this locally and it works very nice! Just one more bug I found, you have a fix for it as a suggestion.

I've tested on U20.04/Noetic and tried searching both Noetic and Jazzy packages (via `--rosdistro`). Both worked as expected.

peci1:
Once this is merged, it should be followed by a flood of tutorial/docs updates :)

StefanFabian:
@cottsay any chance you could review this or assign someone who can do a review?

kscottz:
Thanks for the contribution @StefanFabian 

I ran the tests again to make sure we're good.

Scott is busy with Kilted so I pinged @christophebedard to do a quick second review. Let's get this out the door.

kscottz:
@StefanFabian [looks like an issue with importing regex in the tests.]( https://github.com/ros-infrastructure/rosdep/actions/runs/14378302461/job/41120586348?pr=997)

kscottz:
@StefanFabian This is just my preference but option number 3, " Make regex optional by using a try-catch import and falling back to a search using only re" seems like it has the lowest maintainer overhead with the most cross platform support.

christophebedard:
If it works with the built-in `re` module, I think we should just stick to that and not use/depend on `regex`.

christophebedard:
> If it works with the built-in `re` module, I think we should just stick to that and not (conditionally or not) use/depend on `regex`.

Although I see that fuzzy matching is just not (easily) possible with the built-in `re` module, so yeah let's keep `regex` as an option.

src/rosdep2/main.py
```diff
@@ -918,6 +930,87 @@ def command_resolve(args, options, lookup=None):
...
+    if options.verbose:
+        print("Searching using the following regexes: %s" %regexes)
+    regexes = [re.compile(r, regex_flags) for r in regexes]
+    close_keys = []
+    close_pkgs = []
+    for view_name in sources_loader.get_loadable_views():
+        view = sources_loader.get_source(view_name=view_name)
+        if not isinstance(view, CachedDataSource):
+            if options.verbose:
+                print('Skipping non-cached source %s' % view_name)
```
christophebedard, line 955: Can we tell users what to do in this case?

christophebedard (commented):
Some minor comments.

I think I understand the difference between "keys" and "packages" (i.e., searching actual rosdep keys vs searching what a key might point to), but I think it might be a bit confusing to users if it's not explained. That could be left for some eventual user documentation.

christophebedard:
Oh, and some tests would be great! At least just for the main use-cases.

src/rosdep2/main.py
```diff
@@ -105,6 +114,9 @@ class UsageError(Exception):
 rosdep resolve <rosdeps>
   resolve <rosdeps> to system dependencies
 
+rosdep search <searchstrings>...
+  search for a key in the rosdep database. E.g. using the system package name.
+
```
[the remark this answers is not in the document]
StefanFabian: Are you sure `--rosdistro=` is necessary?
I can find ROS packages just fine without it.

I will make the help string more clear :+1:

src/rosdep2/main.py
StefanFabian: Honestly, I don't know what to do there.
I'm skipping them because I don't know the structure if they're not this type.
This print is purely for debugging purposes.
I think a typical installation wouldn't encounter it, at least I never did.

src/rosdep2/main.py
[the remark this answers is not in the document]
peci1: It gets filled from `ROS_DISTRO` env var.

StefanFabian:
@christophebedard Thank you for the review!
I've addressed your concerns, added tests, and handling of more dependency key structures.

Let me know if there's anything else that needs to be improve.

src/rosdep2/main.py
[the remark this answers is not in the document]
christophebedard: Good point, but it's not available if you haven't sourced ROS (`ros_environment`) in your terminal.

Anyway, like I was saying, this is the case for all commands that need a ROS distro. I just feel like this isn't very clear, but we can always make it clearer later.

src/rosdep2/main.py
```diff
@@ -918,6 +935,127 @@ def command_resolve(args, options, lookup=None):
...
+    close_pkgs = []
+    for view_name in sources_loader.get_loadable_views():
+        view = sources_loader.get_source(view_name=view_name)
+        if not isinstance(view, CachedDataSource):
+            if options.verbose:
+                print('Skipping non-cached source %s' % view_name)
+            continue
+        search_result = search_cached_data_source(view, regexes)
+        close_keys.extend(search_result[0])
+        close_pkgs.extend(search_result[1])
```
christophebedard, line 1020:
```suggestion
        results_keys, results_pkgs = search_cached_data_source(view, regexes)
        close_keys.extend(results_keys)
        close_pkgs.extend(results_pkgs)
```

test/test_rosdep_main.py
```diff
@@ -280,6 +280,34 @@ def test_keys(self):
...
+                assert 'Closest keys' in stdout.getvalue(), stdout.getvalue()
+                assert 'curl' in stdout.getvalue(), stdout.getvalue()
+                assert 'Closest packages' not in stdout.getvalue(), stdout.getvalue()
+                assert not stderr.getvalue(), stderr.getvalue()
+            with fakeout() as b:
+                rosdep_main(['search', 'libeigen3-dev'] + cmd_extras)
+                stdout, stderr = b
+                assert 'Closest keys' not in stdout.getvalue(), stdout.getvalue()
+                assert 'eigen:' in stdout.getvalue(), stdout.getvalue()
+                assert 'Closest packages' in stdout.getvalue(), stdout.getvalue()
```
christophebedard, line 300: I'd just move the `eigen:` line down so it's under `'Closest packages'`.

```suggestion
                assert 'Closest keys' not in stdout.getvalue(), stdout.getvalue()
                assert 'Closest packages' in stdout.getvalue(), stdout.getvalue()
                assert 'eigen:' in stdout.getvalue(), stdout.getvalue()
```

christophebedard:
There are new flake8 errors caused by unnecessary `global` (because the variable doesn't get re-assigned), probably because of a newer flake8 version. They're causing CI to not fully run: https://github.com/ros-infrastructure/rosdep/actions/runs/14750900000/job/41502835987?pr=997#step:5:286. I think we can just remove `global` in this case.

It would be simpler to just fix them in this PR. Could you do that?

src/rosdep2/main.py
```diff
@@ -105,6 +114,14 @@ class UsageError(Exception):
 rosdep resolve <rosdeps>
   resolve <rosdeps> to system dependencies
 
+rosdep search <searchstrings>...
+  Search for a key in the rosdep database.
+  Searches rosdep keys and system or ros package names.
```
christophebedard, line 119:
```suggestion
  Searches rosdep keys and system or ROS package names.
```

src/rosdep2/main.py
```diff
@@ -105,6 +114,14 @@ class UsageError(Exception):
...
+  The search is case-insensitive and all search strings have to match.
+  Supports fuzzy-search if the python module regex is installed.
```
christophebedard, line 121:
```suggestion
  Supports fuzzy-search if the Python module regex is installed.
```

test/test_rosdep_main.py
```diff
@@ -280,6 +280,34 @@ def test_keys(self):
...
+                assert not stderr.getvalue(), stderr.getvalue()
+
```
christophebedard, line 302:
```suggestion
```

src/rosdep2/main.py
```diff
@@ -918,6 +935,127 @@ def command_resolve(args, options, lookup=None):
...
+        item = view.rosdep_data[key]
+        for os_entry in item:
+            os_items = item[os_entry]
+
+            def check_pkgs(items):
+                for pkg in items:
+                    results = [regex.search(pkg) for regex in regexes]
+                    if all(results):
+                        error = count_errors(results)
+                        close_pkgs.append({'key': key, 'pkg': pkg, 'os': os_entry, 'error': error})
```
christophebedard, line 962: I'm a bit confused by the intended or unintended use of the `os` info for packages:

```
$ rosdep search tracetools --rosdistro=rolling
WARNING: ROS_PYTHON_VERSION is unset. Defaulting to 3
Closest keys:
  tracetools_image_pipeline
  tracetools
  tracetools_launch
  tracetools_read
  tracetools_test
  tracetools_trace
  tracetools_acceleration
  tracetools_analysis

Closest packages:
  ros2trace_analysis: ros/rolling/tracetools_analysis [osx]

```

Why is it only listing `osx`? As a user, I would think this means that this package is only available on macOS. Could we instead list all OSes, e.g., `[osx, debian, rhel, ubuntu]`? The code seems to be stopping after finding results for the first available OS.

StefanFabian: I've decided to stop when a match is found because usually it's not relevant if there's another os with a matching name, since you usually know what you are looking for.
The main use case here is also looking for the package name in your distro's package manager to find the associated key, and in that case, only that would match anyway.
Debian and Ubuntu, for example, would almost always both have the same results, and this would unnecessarily bloat the output.
If I were to also list the other OS, it would be confusing if I didn't also list the matching package name for each OS, and this could get quite a long list if the search string matches many packages.

peci1: What about searching just for the currently used OS (or the one passed in `--os`) ?

christophebedard: Yeah, I think that would make a lot of sense. Not sure what the proper way to do it is, but looks like you could use `OsDetect` + the OS override info.

> I've decided to stop when a match is found because usually it's not relevant if there's another os with a matching name, since you usually know what you are looking for.

If the OS information doesn't really matter, it shouldn't be shown IMO.

StefanFabian: Okay, I have changed it to use the os information and only look through packages for the current OS.
This is also updated in the help string, and I've removed the information for which OS the package was found as this is now no longer relevant.

StefanFabian:
I think I've addressed all feedback. If there's something that still needs to be changed, please let me know.

peci1 (approved):
Great, thanks a lot, it works for me with the updates!

christophebedard (approved):
Thanks for iterating! Looks good to me now.

@nuclearsandwich @cottsay I don't have permissions on this repo: could one of you approve the CI and possibly review and/or merge this? Note that this PR includes the same changes as in #1004, since they were necessary to prevent CI from failing fast due to unrelated flake8 errors.

christophebedard:
@StefanFabian can you take a look at the CI failure for the _Run tests / pytest / pytest (macos-latest, 3.12)_ job? https://github.com/ros-infrastructure/rosdep/actions/runs/15075300653/job/42381233133?pr=997

StefanFabian:
@christophebedard I think it was caused by the search command using the os of the test runner.
I've made the os explicit in the tests, so they should run the same on any platform.

christophebedard:
Thanks, that makes sense.

@nuclearsandwich could you approve a new CI run? :sweat_smile:

christophebedard:
CI looks good. Just waiting on an additional review or merge, but people are a bit busy.

christophebedard (approved):

cottsay (approved):
I didn't do a close review here, but I don't see anything dangerous.
Thanks for your contribution!

PR #999, opened 2025-04-26
Don't consider venv as externally managed

krisl:
Allow installation into virtual environments without requiring `--break-system-packages`

### DIFF
--- a/src/rosdep2/platforms/pip.py
+++ b/src/rosdep2/platforms/pip.py
@@ -84,6 +84,10 @@ def get_pip_command():
     return None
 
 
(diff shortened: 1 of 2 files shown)


nuclearsandwich (commented):
Thanks for the PR. Since you contributed a test for this function in #1000 is it worth rebasing now that it has merged and covering the virtualenv case in the test?

krisl:
> Thanks for the PR. Since you contributed a test for this function in #1000 is it worth rebasing now that it has merged and covering the virtualenv case in the test?

cool, I rebased and added a test in there

PR #1000, opened 2025-04-27
Test externally_managed_installable

krisl:
Adds tests for `externally_managed_installable` function

### DIFF
--- a/test/test_rosdep_pip.py
+++ b/test/test_rosdep_pip.py
@@ -83,6 +83,46 @@ def test_PipInstaller_handles_externally_managed_environment(externally_managed_
     assert installer.get_install_command(['whatever'], interactive=False)
 
 
+@patch('sys.version_info', new=(3, 11))
+@patch.dict(os.environ, {'PIP_BREAK_SYSTEM_PACKAGES': '0'})
+def test_externally_managed_installable():
+    from rosdep2.platforms.pip import externally_managed_installable
+    assert externally_managed_installable() is False
+
+    @patch('sys.version_info', new=(3, 10))
+    def test_last_exempt_version():
+        assert externally_managed_installable()
+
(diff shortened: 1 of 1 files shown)


test/test_rosdep_pip.py
```diff
@@ -83,6 +83,46 @@ def test_PipInstaller_handles_externally_managed_environment(externally_managed_
     assert installer.get_install_command(['whatever'], interactive=False)
 
 
+@patch('sys.version_info', new=(3, 11))
+@patch.dict(os.environ, {'PIP_BREAK_SYSTEM_PACKAGES': '0'})
+def test_externally_managed_installable():
+    from rosdep2.platforms.pip import externally_managed_installable
+    assert externally_managed_installable() is False
+
+    @patch('sys.version_info', new=(3, 10))
```
cottsay, line 92: Even though it works to patch the import name on occasion, we should be patching the imported site instead.

See https://docs.python.org/3/library/unittest.mock.html#where-to-patch for details.

```suggestion
@patch('rosdep2.platforms.pip.sys.version_info', new=(3, 11))
@patch.dict(os.environ, {'PIP_BREAK_SYSTEM_PACKAGES': '0'})
def test_externally_managed_installable():
    from rosdep2.platforms.pip import externally_managed_installable
    assert externally_managed_installable() is False

    @patch('rosdep2.platforms.pip.sys.version_info', new=(3, 10))
```

cottsay (changes requested):
Thanks VERY MUCH for adding tests. One comment, then I think this is good to go in.

test/test_rosdep_pip.py
krisl: thanks, addressed

nuclearsandwich (approved):
As the original author of the function under test I echo @cottsay's gratitude. Thanks so much!

PR #1004, opened 2025-05-15
Remove unecessary global declarations.

nuclearsandwich:
Because these variables are in the local scope and are not re-assigned
in these functions, the global declaration can be omitted.

Recent versions of flake8 now raise F824[1] on these lines.
I am not sure if they were added defensively or if they were previously
required to access these variables as globals. Tests all pass as-is but
if we prefer to keep the defensive global declaration we can modify
these lines to `noqa: F824`.

[1]: https://flake8.pycqa.org/en/latest/user/error-codes.html

### DIFF
--- a/src/rosdep2/catkin_packages.py
+++ b/src/rosdep2/catkin_packages.py
@@ -19,7 +19,6 @@ def find_catkin_packages_in(path, verbose=False):
     :returns: a list of packages in a given directory
     :raises: OSError if the path doesn't exist
     """
-    global _catkin_packages_cache
     if not os.path.exists(path):
         raise OSError("given path '{0}' does not exist".format(path))
     if verbose:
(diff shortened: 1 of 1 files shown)


christophebedard (approved):

nuclearsandwich:
Thanks for the lightspeed review @christophebedard. @cottsay do you have any reservations about removing `global` here? I haven't refreshed myself on the Python semantics recently. It's clearly what the linter warning _wants_ but I'm unsure if this opens us up to nastiness, particularly in our older supported Python versions.

cottsay:
> do you have any reservations about removing `global` here?

IMO, it's misleading to have them when they're not needed. Before we removed them from colcon, I was `git blame`-ing my way through the history trying to figure out why they were added and if we missed something. I can see an argument that they serve has a hint for humans reading the code, but I think there are better ways (like appropriate naming) than using `global`.

PR #1009, opened 2025-06-26
Prepare 0.26.0

becschwartz:


cottsay (approved):
:ship:

nuclearsandwich:
@becschwartz the only note I have is that when we do a version bump we make the commit message the new version number exactly. I have updated the PR using `git commit -m '0.26.0' --amend`

